Issue #100, opened 2018-04-12
Question about O.Enrol


FGUERINGTO:
Text of O.Enrol contains some ambiguity. 
The TOE shall implement the functionality to enroll a user for mobile biometric verification, only after successful authentication of the user. The TOE shall check the quality of enrolment templates in order to meet the relevant error rates for mobile biometric verification.

What is the result of the check when quality is good or wrong ?
Could we precise that successful authentication of the user has been done with Fallback mechanism, otherwise it is strange to say authentication is perform prior enrolment.

nils-tekampe:
Closing the Issue as the corresponding PR has just been merged.

PR #114, opened 2018-05-30
O.Enroll update

woodbe:
To address the concern in #100 by @FGUERINGTO

### DIFF
--- a/3_ProtectionProfile/Bio cPP.md
+++ b/3_ProtectionProfile/Bio cPP.md
@@ -411,7 +411,8 @@ rates shall be specified in the FIA\_MBV\_EXT.1.
 
 The TOE shall implement the functionality to enroll a user for mobile
 biometric verification and bind the template to the user only 
-after successful authentication of the user. The TOE shall create the 
+after successful authentication of the user to the TOE environment 
+using an alternative authentication mechanism. The TOE shall create the 
 sufficient quality of templates in order to meet the relevant error 
 rates for mobile biometric verification.
 


stemotre:
Is there a good reason to prevent systems allowing a User to authenticate with a previously enrolled biometric template before enrollment from being compatible with this cPP, and to require them to gate enrollment to an alternative auth method from the environment (i.e. not bio)?

woodbe:
Wouldn't that be a "chicken and egg" problem? How can you enroll the biometric without having a biometric? MDF generally points to requiring a password as the primary authentication method and as such requires the entry of that before being able to change any credentials (at least that is how I read it at this point). So while I can see technically allowing the user to use one biometric to enroll another, I don't think it is what the MDF intends.

Maybe @lewyble can provide a comment on that.

nils-tekampe:
Just a short question from my side: When we talk about "enrolment": Does this exclusively mean the initial enrolment of a user or would an update of an already existing template also be covered by this description?

woodbe:
So I would actually think no, we don't, but as far as the MDF has been concerned, all enrolments are the same, so they always require the entry of the password before you can enroll one. From the MDF standpoint today (and at least how Samsung devices work) is that you always need to enter the password (or PIN/pattern if you aren't in a CC-specific configuration) before you can make any biometric changes (i.e. enroll a template, delete one, etc). So I don't necessarily have an issue allowing a biometric authentication to be allowed to provide sufficient authorization to make a change, but this is different from the current interpretation I have seen from the MDF standpoint.

stemotre:
Apple is currently also requiring Passcode entry before any (first/new/update...) bio enrollment. 
But again, I felt this is not necessarily about our technical choices.
I think leaving this open would allow developers to make their own choice. I understand that choice could "go away" if the developer also complies to MDF PP and this latter requires the main User device password entry.

stemotre:
Ok, this sounds good then.

BrJu:
checked with François: ok

n-kai:
We agree with this PR at the meeting on 7th June.

PR #116, opened 2018-05-31
Update README.md

BrJu:
Add a reference to commit rules


BrJu:
Thanks for the review. This change is minor (not the procedure itself for PRs), but the modification of Readme.md. So if no further comments in the next days, will go for a merge.

Issue #63, opened 2017-11-22
Define asset to protect


woodbe:
From @FGUERINGTO 

In Threats:

As defined in CC §1 A.6.2, “A threat consists of an adverse action performed by a threat agent on an asset”. 
Therefore prior to define threat, it is useful to define assets to protect and associated security property (C,I,A,NR…)

Suggested Changes:

Main assets are : 
Protected Mobile Service [PMS] Security property is Access Control [AC].
	
User identity (e.g. pseudonym) [Confidentiality, Integrity, AUthenticity]
	
Biometric data (Features, samples) [Confidentiality, Integrity, AUthenticity, AVailability]
	
Biometric templates [Confidentiality, Integrity, AUthenticity, AVailability]
	
Fall back Mean (Pin, Password, swipe pattern)
TSF data : 
Decision Threshold [I]

BrJu:
Discussed with François:
-Threshold/biometric parameters integrity still seem important.  Will be good to emphasize the need to protect it even if only one is fixed.
-Importance to discuss the other assets for use case 2+

woodbe:
I would agree this is important to protect, but I don't actually see how the TOE itself would provide this protection as it is defined now. I created #118 to try to handle this case.

@FGUERINGTO please review the PR and comment/approve

BrJu:
Addressed by #118

PR #118, opened 2018-06-05
Biometric configuration protection

woodbe:
This is in response to #63 to add protection of TOE configuration data explicitly to OSP.Protection and OE.Protection.


n-kai:
@woodbe @stemotre will create a new pull request to remove A.management and modify the maping table between assumption and objectives.

woodbe:
I only made a few of the application note numbering changes, but everything needs to be updated as 2 were deleted by removing A. Management (and I didn't update all the ones after that, only the last ones)

BrJu:
Please @n-kai @nils-tekampe have a last look before merging.

BrJu:
Note that application notes numbering will need a new update after App Note 24

Issue #79, opened 2018-03-19
Architecture diagram


woodbe:
https://github.com/nils-tekampe/cPP-biometrics/blob/master/input/architecture.png

I uploaded this diagram responding to the need for some sort of sample architecture referencing a mobile device. Since the mobile devices I know of today do not use external biometric components, I didn't draw that in, I expected everything would be done in a self-contained manner on the device. I didn't give a lot of detail, but some basic points here that should work for the devices I know.

To be clear on it, if you start with the user making a request that needs authentication (an app or the device), then from that starting point, the app (or login service) would call to OS Biometric API layer.

The OS Biometric Layer could be simple or complex, but mainly it is the standardized API provided for the system to access the biometric components. This may also provide some UI elements (such as prompts to the user about the process) or these may be left to the subsystem (i.e. the TOE), depending on the implementation. I show the display hardware inside the SOE (which it usually is for DRM purposes), but that is not a critical component here, just one implementation example.

The OS Biometric Layer then calls to the TOE inside the Secure Operating Environment. I didn't bother to show any actions to the SOE (such as storage) since the TOE is completely contained within the SOE, and relies on its capabilities for storage, processing, crypto, etc.

Once the request to authenticate has reached the TOE, the user would provide their sample, captured by the sensor component of the TOE and perform the processing. The result passed back to the OS Biometric API layer and then to the app/OS to allow or block the user.

Obviously this would be similar for enrollment, etc, I just did it for an auth attempt as the simple version to follow.

woodbe:
I have uploaded my updated image files based on my original file and the changes from @BrJu. I have uploaded a PNG, PDF and the original Visio file to the images folder.

BrJu:
Looks good for use case 1. To be reviewed on meeting 20180607

stemotre:
As exposed in issue #119, the proposed architecture diagram shall discuss plaintext storage in SEE versus secure storage outside SEE (and cPP define “secure" and responsibility of TOE/TSF).
Actually, the SEE boundaries are not relevant to this cPP unless the SEE inherited requirements are clearly addressed and part of this cPP (either requirements or I believe environment objectives).
Attached are two updates that reflect the current state of the cPP (which we have an issue with) and a TOE centric version: with SEE and without.

[...]

PR #121, opened 2018-06-21
Edit to 2.4 for architecture

woodbe:
In reference to #79 to make SEE usage a little more clear


BrJu:
@n-kai @nils-tekampe  can you check quiclky please?

Issue #106, opened 2018-05-17
BS SD (EA for FPT_BDP_EXT.1)


The-Fiona:
The statement "However, the evaluator shall examine that the TOE processes any plaintext biometric data except publicly accessible biometric data (e.g. face image) within the security boundary of the SEE." is causing me to ask a question:

How do we define what is publicly accessible biometric data ?
When I go out in public usually my face is available for all to see, but I also leave my fingerprints, and samples of DNA around (sorry for that thought!).  I also understand that fingerprints may be "readable" by cameras nowadays, and have read some articles on long-range iris-scanning.
So, assuming that this individual does go out in public, is there any biometric data that is not publicly available?

So, in this EA I guess we either need to precise "publicly accessible biometric data" or remove the exception - Which would effectively mean that we ask for all biometric data to be stored in the SEE.

Welcoming comments on this matter :)

woodbe:
@n-kai 

So generally when a sensor will be used for authentication (especially for MDF-targeted devices) the sensor will be incorporated into the boundary of (for lack of a better descriptive term) the Dedicated Security Components (DSC). So in the case of a Samsung device, the fingerprint sensor has a dedicated path to the TEE and cannot be accessed by the Rich OS. Generally all authentication mechanisms would be targeted in this way so that the core logic/processing would occur in the SEE (whatever that is), whether it is a password or biometric being used to authenticate the user. 

So for something like a front-facing camera which may be configured to be able to be used both as an authentication sensor and as a selfie camera, the camera would normally be plugged into the DSC boundary, but have the option that if it is not being used for authentication purposes to allow an application to request access to the image produced by the camera. This is what I would expect, anyway. I have not looked into this specifically on Samsung devices or Android in general as we have not tried to validate the face unlock function (we have limited ourselves to fingerprint and iris to date).

To that end though, if the sensor has the ability to be used by the Rich OS and the SEE, it must support some sort of priority-based exclusive access lock so that if the sensor is being used for an authentication operation by the SEE that nothing else can gain access until that operation is completed (and probably for some sort period right after to prevent say taking the picture of a face right after successful authentication since that would be a great time to grab the image).

n-kai:
@The-Fiona @woodbe @BrJu 
Thanks for your input. I agree with the proposal from @BrJu in #85 to remove "except publicaly accessible biometric data" from the requirement.

n-kai:
[...]

Issue #85, opened 2018-04-05
change request to the cPP


n-kai:
I have the following change proposals to the latest cPP to sync it with the latest SD (https://github.com/nils-tekampe/cPP-biometrics/blob/NKAI-April-05/input/BS%20SD.md)

**CP1**
Replace
_FPT_BDP_EXT.1 Biometric data processing
FPT_BDP_EXT.1.1 The TSF shall process any plaintext biometric data used to generate templates and perform sample matching, except publicly accessible biometric data, within the security boundary of the secure execution environment._

_FPT_BDP_EXT.3 Biometric data storage
FPT_BDP_EXT.3.1 The TSF shall not store any plaintext biometric data except publicly accessible biometric data outside the secure execution environment._

with
_FPT_BDP_EXT.1 Biometric data processing
FPT_BDP_EXT.1.1 The TSF shall process any plaintext biometric data used to generate templates and perform sample matching, except [assignment: list of publicly accessible biometric data], within the security boundary of the secure execution environment._

_FPT_BDP_EXT.3 Biometric data storage
FPT_BDP_EXT.3.1 The TSF shall not store any plaintext biometric data except [assignment: list of publicly accessible biometric data] outside the secure execution environment._

**Rationale**
“publicly accessible biometric data” should be defined accurately by the ST author.  

**CP2**
Replace
_FPT_BDP_EXT.2 No Biometric data transmission
FPT_BDP_EXT.2.1 The TSF shall not transmit any plaintext biometric data except publicly accessible biometric data outside the security boundary of the secure execution environment._

with
_FPT_BDP_EXT.2 No Biometric data transmission
FPT_BDP_EXT.2.1 The TSF shall not transmit any plaintext biometric data outside the security boundary of the secure execution environment._

**Rationale**
Biometric data is sensitive personal data and publicly accessible biometric data (e.g. face image) shall not be transmitted to any entities outside the SEE without a consent from users.

BrJu:
"My opinion is that anything that is taken by the TOE for the purpose of generating a biometric template or as a sample to match to a template is not really "public" data since it should only be available inside the TOE. As such, I would remove the "except publicaly accessible biometric data" from the original one as my suggested modification."

I support this removal. The TOE should not make any distinction between the processed biometric data.

BrJu:
cf. changed made in #118 for FPT_BDP_EXT.3.1. 
(the use of "except publicaly accessible biometric data" should be checked elsewhere in the cPP)

[...]

PR #122, opened 2018-06-21
Remove publicly accessible biometric data

woodbe:
Proposal to remove "publicly accessible biometric data" from the cPP.

#85 and #106

@n-kai please review. I mainly just removed that text, I didn't see it needed to be further modified.


woodbe:
As a note, this removes several Application Notes which will have some impact on their numbering, but until the #118 has been merged and closed, the final numbering is not completely clear.

BrJu:
Thanks @woodbe. Please @n-kai tell us if any comments

PR #125, opened 2018-07-12
application note numbering fix

woodbe:
This is only to fix the renumbering of the application notes. There were several merges that had different impacts on the numbering. Now that those were merged, this was to review the updated document and correct the numbering on the application notes. No other changes.


BrJu:
to be merged with #127

PR #129, opened 2018-07-19
FIA_HYB_EXT edits

woodbe:
The description said a PIN or password and MDF requires a password, so listing PIN isn't a valid configuration. 

I also found a lowercase "if" in the last sentence.

I fixed these in both places.


BrJu:
@woodbe we can see several references to PIN or password (including for the hybrid mechanism) in MDFPP v3.1 https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html 

Could you check if we are not refering to the right version?

woodbe:
My mistake, looking at it the PIN is there for hybrid (but not stand alone, which is what I was thinking about). I have updated the PR to only be for the typo.

Issue #128, opened 2018-07-19
A.Alternative include mention of enrolment?


woodbe:
Should A.Alternative have a mention that it is also required for enrolment of a biometric and not just as a "fallback" method if the biometric fails?

> It is assumed that the TOE environment provides an alternative authentication mechanism as a complement to mobile biometric verification. _The alternative authentication mechanism is required for enrolment of the biometric template and can also_ be used in cases when a user is rejected by the mobile biometric verification (False Rejection).

The italics is a possible change (to the second sentence).

A similar edit can be made to OE.Alternative as well.

nils-tekampe:
I have never seen it this way. From my perspective: If you fail enrolment (whyever) and keep failing it, you will have to use the fallback every time you come to the biometric system because you know that you will be failing to use the biometric authentication. 

But I can see your point and would support the change.

woodbe:
@nils-tekampe I agree with your perspective, but as I was actually reading through from start the end, I came across the earlier descriptions and when I got to this it occurred to me that may it should be noted that it is the same mechanism that is required for the initial enrolment as well.

I am not wedded to making this change, but I thought that it brought some clarity as well as tightened up where the alternative authentication mechanism is used.

BrJu:
That's a useful clarification.

BrJu:
Closing as addressed by #130

PR #130, opened 2018-07-19
A.Alternative suggestion

woodbe:
This is to provide a full change suggestion for #128.


BrJu:
To be validated during Aug 2nd meeting

nils-tekampe:
Discussed during the call and merged without objections.

PR #134, opened 2018-08-01
App Note numbering

BrJu:
correcting again...


nils-tekampe:
Merged as discussed during the call

PR #136, opened 2018-08-23
N kai sd patch 201808

n-kai:


nils-tekampe:
I just started editing the contribution and started with chapter 5 (the assurance activities). In my opinion we should try to refer more extensively to ISO/IEC 19989-1 in this chapter. I think that many of the aspects that have been in these application notes are already covered by the standard. So, I removed many of the app notes and added a reference to 19989-1. I left an application note that I think is really specific to the cPP. I'm really looking forward to your feedback on this. 
If we should find that the one or the other detail may not be covered by 19988-1 yet, we also should consider to feed the content into ISO.

n-kai:
Publishing the SD at the CCRA portal was what the Certification Bodies(CB) agreed at the CCRA meeting when it started the cPP/SD based evaluation.

There is another reason. The CCRA/CB is responsible for maintaining the evaluation standards/methodology like CC/CEM or cPP/SD. Anyone can submit a change proposal to the evaluation standards etc through the CB. But what happens if someone find a error in 19989-1 which our SD refers and submit the change proposal to the CB?  The CB can propose the ISO to revise the 19989-1 but such revision request may be rejected based on the result of NB vote, which means that the CB/CCRA can't fix the error in the SD.

19989 is still useful document for the evaluation of generic biometric verification and identification. However 19989 doesn't specify any specific activities for mobile biometric evaluation based on our cPP. Such specific information should be described in the SD.

nils-tekampe:
I‘m just wondering whether it would change anything if we would convince ISO to make 19989-1 available for free (as they do with 15408). What do you think?
Also, I would say that if a demand for a change should come up later in the process and the change would be rejected by ISO (for what reason ever) we still can change it in the SD directly. I know many spedicifcations that say something like: „Please follow Standard XYZ except for the following..“. 
@n-kai The agreement at the CCRA meeting that you are talking about. Can I read this anywhere?

n-kai:
Making 19989-1 available for free doesn't solve the problem. For example, all ISO documents need to go through systematic review every 3 or 5 years and someone may change or remove the parts in 19989-1 the SD refers in the future. SD should include all necessary information in 19989-1 so that we can avoid such risk.

>we still can change it in the SD directly.
This means that you propose to copy relevant parts in 19989-1 to the SD so that we can fix the error by ourselves? If so, I agree with you and please tell me which parts in 19989-1 should be copied to the SD. 19989-1 doesn't include any specific information for mobile biometric security but I will try to update the SD as you suggest (or you can update your PR by yourself).

>Can I read this anywhere?
I am not sure where such information is recorded but, for example, iTC process paper available at the CCRA portal requires the iTC to submit the SD/cPP for public review. We can not post 19989-1 for such public review.

nils-tekampe:
As discussed during todays meeting, I'm merging this branch so that we can use it as a starting point for our further work.

PR #139, opened 2018-11-21
First pass on iris toolbox

woodbe:
Updated files to complete a first pass on the iris toolbox.

I modified some of the files within the Methodology folder to better describe 2D image PAD. From there I created a full subfolder for iris. I copied the original files from the face detection and modified them for iris.


woodbe:
In some of the documents there is text that is _**bold italics**_. These are comments I made while editing, and I can bring them into here once the PR is approved.

I am requesting a quick approval here so we can comment/work on the individual issues that may be here instead of having to look specifically at this very large, monolithic set of documents (2 modified files and 22 new ones).

PR #144, opened 2018-11-30
First pass on vein toolbox

n-kai:
Please approve this PR quickly so we can comment/work on any issues in the toolbox as Brain also requested in PR #139 (See below)

_I am requesting a quick approval here so we can comment/work on the individual issues that may be here instead of having to look specifically at this very large, monolithic set of documents (2 modified files and 22 new ones)._


n-kai:
You can view the vein toolbox directly from 
https://github.com/biometricITC/cPP-biometrics/tree/NK-veintoolbox/4_Methodology/attacks_for_vein

n-kai:
@stemotre 
" _I am concerned that the presentations in the tests are only going to prove that the authentication system does not work on photos/images, rather than proving the correlation of a hand and the presented vein schematics_"

I assume that the vein toolbox is used for ATE_IND.1 (evaluator independent testing) without any changes and for AVA_VAN.1 (penetration testing) with some changes. Evalauation activivities for ATE_IND.1 defined in CC/CEM don't fit for this kind of testing and don't include useful information for developing the toolbox. So I created the toolbox based on AVA_VAN.1 evaluation activities that require evaluator to devise tests based on only publicly available information, such as research papers and all english and japanese research papers use photo image to create the vein PAI. That's the reason why the toolbox doesn't include tests you suggested (however, evaluator shall cosider internal information such as functional spec at AVA_VAN.2 and above to devise test items. so such tests may be necessary for AVA_VAN.2 and above evaluations)

PR #147, opened 2019-01-12
Update Bio cPP.md

nils-tekampe:
Translated the cPP to Asciidoctor format. The content is unchanged.


nils-tekampe:
Merged as discussed during the meeting. Will be used for further work then

PR #151, opened 2019-02-18
PR for uploading PP conf and SD in Asciidoc format and update the cPP module

n-kai:
I added two adoc files for PP configuration and SD. I also corrected some editorial errors in the cPP module.
Please simply approve this PR so that we can disscuss based on those latest version.

### DIFF
--- a/3_ProtectionProfile/PP config.adoc
+++ b/3_ProtectionProfile/PP config.adoc
@@ -1,0 +1,121 @@
+= PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device -
+:showtitle:
+:toc:
+:table-caption: Table
+:revnumber: 0.8
+:revdate: 2019-01-31
+
+== Acknowledgements
+
+This PP-Configuration was developed by the Biometrics Security (BS) international Technical Community (iTC) with representatives from industry, Government agencies, Common Criteria Test Laboratories, and members of academia.
+
+:sectnums:
+
+== Introduction
+
+=== Overview
+
+The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that combines Protection Profiles (PPs) and PP-Modules for various technology types into a single configuration that can be evaluated as a whole. The scope includes the definition of the configuration of a mobile device that has biometric enrolment and verification capability. The TOE will be defined by a combination of the components described in section 1.3.
+
+=== PP-Configuration Reference
+
+This PP-Configuration is identified as follows:
+
+* PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, January 31, 2019
+* As a shorthand reference, it can be identified as "BIO+MDF"
+
+=== PP-Configuration Components Statement
+
+This PP-Configuration includes the following components:
+
+* Base-PP: Protection Profile for Mobile Device Fundamentals [MDFPP]
+* PP-Module: collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device - [BIOcPP Module].
+
+== Conformance Claims
+
+=== CC Conformance
+
+*Conformance Statement*
+
+To be conformant to this PP-Configuration, an ST must demonstrate Exact Conformance, a subset of Strict Conformance as defined in [CC1] (ASE_CCL).
+
+The ST must include all components in the Base-PP that are:
+
+* Unconditional (which are always required)
+* Selection-based (which are required when certain selections are chosen in the unconditional requirements)
+
+and may include components that are
+
+* Optional
+* Objective.
+
+The same conditions apply to [BIOcPP Module] that is included as part of this PP-Configuration.
+
+*CC Conformance Claims*
+
+This PP-Configuration, [MDFPP] and [BIOcPP Module] are conformant to Common Criteria Version 3.1, Revision 5.
+
+=== SAR Statement
+
+In order to evaluate a TOE that claims conformance to this PP-Configuration, the evaluator shall evaluate the TOE against the following SARs that are defined in the [MDFPP]:
+
+[cols=",",options="header",]
+.Assurance Components
+|===
+|*Assurance Class* |*Assurance Components*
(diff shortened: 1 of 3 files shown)


3_ProtectionProfile/PP config.adoc
```diff
@@ -0,0 +1,121 @@
...
+|[CEM] |Common Methodology for Information Technology Security Evaluation, +
+Evaluation Methodology, +
+CCMB-2017-04-004, Version 3.1 Revision 5, April 2017.
+|===
+
+*Protection Profiles*
+
+[cols=",",]
+|===
+|[MDFPP] |Protection Profile for Mobile Device Fundamentals, Version:3.1, 2017-06-16
```
The-Fiona: If it is the current 3.1 MDF then it cites CC V 3.1 R4.. It is the candidate MDF 3.2 that cites CC 3.1 R5

3_ProtectionProfile/PP config.adoc
```diff
@@ -0,0 +1,121 @@
...
+Note that to fully evaluate the TOE, these SARs shall be applied to the entire TSF and not just the portions described by [MDFPP] where the SARs are defined.
+
+In addition to this, both [MDFPP] and [BIOcPP Module] define "Evaluation Activities" for how to evaluate individual SFRs as they relate to the SARs for ASE_TSS.1, AGD_OPE.1, and ATE_IND.1. [MDFPP] and [BIOcPP Module] also provide Evaluation Activities for the SARs. In evaluating this PP-Configuration, the evaluator shall ensure that all Evaluation Activities for SFRs and SARs are evaluated as part of satisfying the required SARs.
+
+=== Related Documents
+
+**Common Criteria**footnote:[For details see http://www.commoncriteriaportal.org/]
+
+[cols=",",]
+|===
```
The-Fiona: We should cite the current addenda here.
Here I am assuming that we match the current draft of MDF (Candidate v 3.2) https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.pdf

3_ProtectionProfile/PP config.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Base-PP: Protection Profile for Mobile Device Fundamentals [MDFPP]
+* PP-Module: collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device - [BIOcPP Module].
+
+== Conformance Claims
+
+=== CC Conformance
+
+*Conformance Statement*
+
+To be conformant to this PP-Configuration, an ST must demonstrate Exact Conformance, a subset of Strict Conformance as defined in [CC1] (ASE_CCL).
```
The-Fiona: Sadly, neither [CC1} Not the addenda define the term "Exact Conformance".
Noting that many of the cPP "get around" this issue by defining it within the cPP.
The candidate MDF PP V3.2 is incorrect with the statement "An ST must claim exact conformance to this PP, as defined in the CC and CEM addenda for Exact Conformance, Selection-Based SFRs, and Optional SFRs (dated May 2017)." 

So I would suggest that this PP-Module avoid the issue for now and change the text to:
"To be conformant to this PP-Configuration, an ST must demonstrate Exact Conformance, a subset of Strict Conformance as defined in [MDFPP]."

3_ProtectionProfile/PP config.adoc
```diff
@@ -0,0 +1,121 @@
...
+| |ST Introduction (ASE_INT.1)
+| |Security Objectives for the Operational Environment (ASE_OBJ.1)
+| |Stated Security Requirements (ASE_REQ.1)
+| |Security Problem Definition (ASE_SPD.1)
+| |TOE Summary Specification (ASE_TSS.1)
+|Development (ADV) |Basic Functional Specification (ADV_FSP.1)
+|Guidance Documents (AGD) |Operational User Guidance (AGD_OPE.1)
+| |Preparative Procedures (AGD_PRE.1)
+|Life Cycle Support (ALC) |Labeling of the TOE (ALC_CMC.1)
+| |TOE CM Coverage (ALC_CMS.1)
```
The-Fiona: Add ALC_TSU_EXT

The-Fiona:
Dear Naruki-san,
Sorry I am not very well familiar with the commenting/approving process :(
I did have some suggested changes (That I hope you can see.) But have approved the PR as you originally requested.

Does that work?

- Fiona

n-kai:
@The-Fiona 
Thank you for your review and comment!
I will create another Pull Request to modify the draft as you suggested.
Naruki

woodbe:
I have started a new PR #155 for updates on this which I expect to finish tomorrow (mainly small changes).

There are two points though that I would make:
1. For references in the document, wherever possible it should just use the <<...>> linking method instead of listing the "Section X" in the reference. I changed several of these because it will make editing easier over the long term since you wouldn't have to go back and edit all those references if the section numbers were changed due to larger document changes. There is another one on line 270 that I haven't yet figured out how to change the text for.
2. The beginning should be edited generally to better match the cPP. What I mean here is the ordering of the content. Things like the Glossary and Acronyms should be in the very first sections of the document as they are in the cPP. I wasn't ready for this more wholesale edit right now, but is something we should handle before the end of the month.

woodbe:
@nils-tekampe I agree. Do you have specific suggestions based on the current text that you could put here?

3_ProtectionProfile/PP config.adoc
```diff
@@ -0,0 +1,121 @@
...
+|Tests (ATE) |Independent testing - conformance (ATE_IND.1)
+|Vulnerability Assessment (AVA) |Vulnerability Survey (AVA_VAN.1)
```
nils-tekampe: This line will have to be removed

nils-tekampe:
Yes, at least kind of. First, we need to remove the line where the assurance component AVA_VAN is used. I just put a comment on this line. 
However, this is only the first step. As far as I understand it, it will not be possible to do this within a PP module while the whole PR currently makes the PP from a standalone PP into a module. 

So, my suggestion would be to 
1) remove AVA_VAN, 
2) go thru every change but only accept those ones that do not turn the PP into a module.

woodbe:
Based on general agreement at this time (with @nils-tekampe not in favor) on the 3/14/19 call these changes will be accepted, including leaving AVA_VAN in the cPP-module.

There is still debate about whether this will be the final, and new issues may be opened to this question.

PR #161, opened 2019-05-01
Update BiocPP.adoc

woodbe:
Changed cPP Module to PP-Module related to #160


woodbe:
PP-Module isn't (which I could add), though I'm not sure it is required in this case. 

@nils-tekampe The only change here was to clarify the terms of what we currently have. I do understand you think the PP approach would be better, but for right now, I think we need to get this out for public review, and things like this should follow the expected format (and right now we are a PP-Module, not a PP). 

If we determine we need to change based on the scheme discussions, then I don't have an issue with this all being changed, but that is a small amount of content, both to make the boilerplate change to the PP and the AVA_VAN issues (which have been noted in the review document). In the meantime though, I would like to publish this as close to being ready for publication as possible, and this does seem to be something to clean up.

BrJu:
Trying to catch up with those discussions. 
It seems there are concerns on both side that are difficult to address. Have you ever discussed about going into public review with both options cPP and module, explaining the issue, and to ask experts to comment on that specific issue?

woodbe:
Merged based on 6/13 call.

PR #162, opened 2019-05-01
Update PP config.adoc

woodbe:
Changed title and date to match PP-Module based on #160

### DIFF
--- a/3_ProtectionProfile/PP config.adoc
+++ b/3_ProtectionProfile/PP config.adoc
@@ -7,7 +7,7 @@
 
 == Acknowledgements
 
-This PP-Configuration was developed by the Biometrics Security (BS) international Technical Community (iTC) with representatives from industry, Government agencies, Common Criteria Test Laboratories, and members of academia.
+This PP-Configuration was developed by the Biometrics Security international Technical Community (BIO-iTC) with representatives from industry, Government agencies, Common Criteria Test Laboratories, and members of academia.
 
 :sectnums:
 
--- a/3_ProtectionProfile/PP config.adoc
+++ b/3_ProtectionProfile/PP config.adoc
@@ -22,13 +22,13 @@ The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that
 This PP-Configuration is identified as follows:
 
 * PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative PP-Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, May 01, 2019
-* As a shorthand reference, it can be identified as "BIO+MDF"
+* As a shorthand reference, it can be identified as "PPC+MDF+BIO+01"
 
 === PP-Configuration Components Statement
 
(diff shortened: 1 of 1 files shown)


3_ProtectionProfile/PP config.adoc
```diff
@@ -21,15 +21,15 @@ The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that
...
 
-* PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, January 31, 2019
+* PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative PP-Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, May 01, 2019
 * As a shorthand reference, it can be identified as "BIO+MDF"
 
 === PP-Configuration Components Statement
 
 This PP-Configuration includes the following components:
 
 * Base-PP: Protection Profile for Mobile Device Fundamentals [MDFPP]
```
The-Fiona: The latest term agreed is "base PP"  (lower case and no hyphen)

3_ProtectionProfile/PP config.adoc
```diff
@@ -21,15 +21,15 @@ The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that
 
 This PP-Configuration is identified as follows:
 
-* PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative Protection Profile Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, January 31, 2019
+* PP-Configuration for Protection Profile for Mobile Device Fundamentals and collaborative PP-Module for Mobile biometric enrolment and verification - for unlocking the device - version 0.8, May 01, 2019
 * As a shorthand reference, it can be identified as "BIO+MDF"
```
The-Fiona: I would suggest that
a) Add a prefix to identify this as a PP-Configuration. If we can establish such a convention it will assist schemes / CCDB determine the type of document when they add it to their databases
b) Potential add a number at the end, since in theory we may have more than one  PP-Configurations that are similar - So, for example either the module or MDFPP may rev, and so we would need another PP-Configuration for each revision.

So, I suggest "PPC+BIO+MDF+01"

3_ProtectionProfile/PP config.adoc
woodbe: I will check all the other docs for this (I have been writing it like this for some time).

woodbe: This has been completed (for all docs)

3_ProtectionProfile/PP config.adoc
woodbe: I have edited the short name but used PPC+MDF+BIO+01 instead so the base PP would be first and the PP-Module(s) would come after. I thought precedence of base to Module should be shown in the PP-Config name (as it currently is in the title as well).

PR #164, opened 2019-05-01
Public review pp module update

woodbe:
Covers comments from @The-Fiona and #160


woodbe:
@nils-tekampe Do you think we need to mention something in this document about the possibility of becoming a PP instead of a PP-Module? I don't think that is necessary, but it could easily be added.

woodbe:
My last commit has a LOT of changes to the content. I added a diagram and timeline for the release (which if it was followed would take about 5 months from the date we publish it for public review to be complete), so please take a look at that as well. I also added a description about the question of PP vs PP-Module at the end to be something we would take comments on.

nils-tekampe:
I just made some minor changes to the document. Also, I added an excel-Sheet for the comments

And I'd like to raise one question: Are we (meaning the itc) sure to have sufficient ressources to meet the schedule that is described in the document?

woodbe:
[Comment_Matrix-PP_Module.xlsx](https://github.com/biometricITC/cPP-biometrics/files/3141971/Comment_Matrix-PP_Module.xlsx)

I'm just attaching it here so it can be seen, but I was basically going to crib the NIAP comment matrix, I just hadn't uploaded it. 

I like the changes @nils-tekampe made to my updates

woodbe:
Now as to the question about resources to meet the schedule in the document, I don't know (given how few of us there are), but I'm not really betting on a lot of comments from outside (if there are, well then we could be forced to adjust the timelines, just like we are now waiting for the schemes). 

In my experience making edits is certainly faster than writing new, so I think this is doable. My main concern is if we say double it, then we may not actually have a completed set of docs in 2019.

Of course depending on what guidance we get on the AVA_VAN may throw everything off anyway. So I'm open to longer times if necessary, but I would rather not start with really long ones (because I'm also worried about maintaining what little momentum we have moving here).

woodbe:
I wasn't going to bother with the GPOS PP-Configuration at this point. I think we already have enough to worry about immediately and the PP-Configuration doesn't have much to it.

woodbe:
I'm good with the changes adding the PP-Configuration to the document as it is.

woodbe:
Brian will add toolbox document to the review list and add explicit note about number of user/subjects for PAI creation

4_Methodology/Public_review.adoc
```diff
@@ -1,57 +1,148 @@
...
+==== Spreadsheet Review Process
 Along with the link in <<DocTable>> for each document is a link to the spreadsheet to be used for comments. Email the spreadsheets to <some email address>.
 
+== Additional Notes
+The documents under review are the current state of output. Two major areas of discussion within the iTC that are still under consideration within are noted below. 
+
+Comments on these specific topics can be made separately.
+
+=== Publish as PP-Module or PP
+The original focus for the iTC was to create a standalone PP that would be evaluated separately that defines security requirements to provide the secure environment for the biometric enrolment and verification. Later the iTC modified the PP into a PP-Module for somewhat easier integration with the MDFPP (the primary target device at this time).
```
The-Fiona: "(the primary target device at this time)" :> "(the primary base PP at this time)"

woodbe: I have made this edit

woodbe:
I removed the comments template from this PR in favor of the one from #168 instead

Issue #143, opened 2018-11-26
Motivation description


woodbe:
I don't have a good answer for this, but the motivation has long been a problem for me. Stating that we are only testing against friends playing around (but yet they are willing to spend money or use potentially expensive equipment to do so) I think is a little disingenuous. Not to mention I'm not sure how that really fits with the MDFPP, which doesn't protect necessarily against full-on nation-state level of attacks, is certainly more secure than "my friend is trying to get into my phone for fun".

While I can understand the potential distinction between attacks based on money, I'm not sure there really is a sufficient difference between accessing money and accessing data (or other services/apps, etc) protected by a biometric login (the main one seems to be who has to deal with the consequences of the bypass, not the level of security needed on the access control itself). I can understand different groups who may rely on the cPP for an evaluation needing different assurance, and as had been discussed in Amsterdam, I think that the real difference between level 1 and 2 will largely be about the level of testing (maybe the addition of AVA_VAN.1 for the level 2?), but that the threat itself is really the same, and the motivation is the same (it is not someone hacking for fun, they want access to what is on the device).

I think we should look at rewriting the motivation descriptions we are using to that end. I don't have specific suggested text at this time, but would at a minimum remove the "attacking a friend's device for fun" part of any description as I do not think that is a real-world threat that we would be investing time and effort for a CC evaluation into.

woodbe:
@n-kai have you looked at how to change this regarding the SD? The only place I see this really is [6.3.1.2.2](https://github.com/biometricITC/cPP-biometrics/blob/master/4_Methodology/BS_SD.adoc#63122-new-pais-found).

If we are going to send this out for public review soon, I would like to take care of this if we can.

n-kai:
@woodbe 
The following is my proposed change so pls tell me it's fine for you. I modified [6.3.1.2.2](https://github.com/biometricITC/cPP-biometrics/blob/master/4_Methodology/BS_SD.adoc#63122-new-pais-found) referring [3.5.1](https://github.com/biometricITC/cPP-biometrics/blob/master/3_ProtectionProfile/BiocPP.adoc#351-use-case-1-mobile-biometric-verification-for-unlocking-the-mobile-device)

I will crate the PR to implement the change if it's fine for you.

a. Attacker’s motivation
[...]

PR #165, opened 2019-05-15
Update BS_SD.adoc for #143

n-kai:
The PR for #143 (Motivation description)


woodbe:
Brian will suggest a change to the last sentence in the first paragraph that the SD assumes Use Case 1

woodbe:
Merged based on approval after 5/30 call.

PR #167, opened 2019-05-16
minor update to BS_SD for section 8.1

woodbe:
The sentence was a little hard to read so I added an extra "to" to try and help make it a little more readable. I also added the period at the end which was missing. No other changes


woodbe:
closed based on 5/30 call.

Issue #169, opened 2019-06-03
Change ISO21879 as it is still in draft


woodbe:
From @ohenniger :

ISO/IEC 21879 (under development) is referenced to suggest a FAR target value. Better reference a finished document dealing with this issue.  

Replace [ISO21879] with "[ISO/IEC TR 29156] Information technology -- Guidance for specifying performance requirements to meet security and usability needs in applications using biometrics. 2015"

woodbe:
Note that I have created PR #170 which changes ISO21879 to ISO19795-9 in reference to a different comment, but it is still under draft and could be further changed.

woodbe:
Second, related comment on changing to the ISO29156:

ISO/IEC PDTS 21879 became ISO/IEC DTS 19795-9. ISO/IEC DTS 19795-9 does not mention "four-digit PIN code". 

Replace 
"[ISO21879] is proposing that FAR would be 1 in 10000 or lower that is equal to a conventional four-digit PIN-Code for secure transaction" 

with 

"[ISO/IEC TR 29156] suggests as a simple rule of thumb that for basic, medium and high levels of authentication assurance, rates of 1% (1 in 100), 0.01% (1 in 10^4) and 0.0001% (1 in 10^6) can be considered as suitable target figures for FAR".

woodbe:
I have proposed this change in #175. We could close #170 in favor of this change instead.

woodbe:
This has been added to the latest version.

PR #170, opened 2019-06-03
edit ISO21879

woodbe:
Updated reference for ISO21879 to ISO19795-9


woodbe:
This is based on comments from @ohenniger

n-kai:
I agree with comments from @ohenniger in #169. However according to #169, the following changes should be made to the draft but this PR proposes different updates. Am I looking at the wrong PR?

**Replace [ISO21879] with "[ISO/IEC TR 29156] Information technology -- Guidance for specifying performance requirements to meet security and usability needs in applications using biometrics. 2015"**

**Replace
"[ISO21879] is proposing that FAR would be 1 in 10000 or lower that is equal to a conventional four-digit PIN-Code for secure transaction"
with
"[ISO/IEC TR 29156] suggests as a simple rule of thumb that for basic, medium and high levels of authentication assurance, rates of 1% (1 in 100), 0.01% (1 in 10^4) and 0.0001% (1 in 10^6) can be considered as suitable target figures for FAR".**

woodbe:
So there were several changes recommended around this and I wasn't sure how many of them to implement. I made the "basic" change of the ISO21879 to ISO19795-9, but there were then suggestions to change to ISO29156 instead. I only made the first change, which could be rejected in favor of the broader changes to the 29156 change instead, but I was not as certain about those references.

PR #174, opened 2019-06-12
Updated Public Review

woodbe:
Updated email addresses, document links, dates. Added separate table for spreadsheet (used a table so if in the future more than one supporting document is provided they can easily be listed without major format changes).

All links currently point to CCUF OO portal in this document.


woodbe:
I realized the URLs to OnlyOffice are wrong, and I can't create them until I upload the document to the site. That said, I'm wondering about it being where it requires a login to get to it, so I'm wondering if we should stick to having it on the GitHub link as the primary, and then inside OO post a forum notice that points to both locations with the OO one being first.

@The-Fiona what do you think?

woodbe:
Merging based on 6/13 call

PR #175, opened 2019-06-13
ISO 29156 edit

woodbe:
alternate proposed change to #170 .

In this one the draft ISO (21879 or 19795-9) is replaced with 29156. This also includes the suggested text edits supporting the FAR values from the ISO.


woodbe:
Brian will merge this in based on 6/13 call.

Pointed out that the references can always be changed during the review period.

Issue #173, opened 2019-06-03
Update relationship to other ISO biometric standard under development


woodbe:
From @ohenniger:

BiocPP and BS_SD do not reference the following related standards documents under development:      

[ISO/IEC 19989-2] Information technology -- Security techniques -- Criteria   and methodology for security evaluation of biometric systems -- Part 2: Biometric   recognition performance     

[ISO/IEC 19989-3] Information technology -- Security techniques -- Criteria   and methodology for security evaluation of biometric systems -- Part 3:   Presentation attack detection     

[ISO/IEC 19795-9] Information technology -- Biometric performance testing   and reporting -- Part 9: Testing on mobile devices     

[ISO/IEC 30107-4] Information technology -- Biometric presentation attack   detection -- Part 4: Profile for testing of mobile devices

Clarify the relationship to these documents

n-kai:
Yes, those references should be removed from the draft. 
I will do a final check to find other errors in the draft before the public review.

woodbe:
So thinking about the change in #175. I assume it is probably worth while to have these all listed as reference documents, so that they are all listed as related but I don't know that we need to go any further. This would also bring back the drafts (now 21879) as related, but leave the change in #175 as the one that is referenced later in the document. This may actually be a more clear point since these are all related. I created #176 to do this.

woodbe:
This has been added to the latest release.

PR #176, opened 2019-06-13
Added additional ISO references

woodbe:
Based on #173 I have added all the other references to the document.

While I removed the 21879 in favor of the 29156 earlier, this update brings it back just as a reference.


woodbe:
Brian will edit the ISO listings to either have a "/" or "_" based on internal referencing after the call.

PR #184, opened 2019-08-05
Public Review 1 update for SD

woodbe:
Changes from @n-kai

Changed admonition style to better supported version


woodbe:
I could not see any way to preserve the list flow while moving that to the left, so I moved it above the next list item (so it could move to the left) and then continued the list. So now that text is on the left margin, but it is above b) but I think it should work.

Issue #187, opened 2019-09-03
Edit PP-Module to not be MDFPP-specific


woodbe:
I am wondering if we should consider editing the PP-Module to move things like the MDFPP justifications to somewhere else (I'm wondering if these should actually be in the PP-Configuration since that actually makes more sense as the place to justify why the PP and PP-Module would go together).

I do not know if this is actually possible in terms of the expectations of the PP-Configuration document, but I am thinking here that the document would be a little more "generic" in that it relies on base-PP, but not a specific one, and that the justification which PPs it can work with should in the PP-Configuration.

Is this workable within the expectations of the PP-Configuration?

n-kai:
I agree with you. 

Current CC/CEM V3.1.5 is not clear about it but according to the latest ISO version of CC (ISO/IEC 15408-1 CD3);

_C.3.1.5 Consistency rationale
A PP-Configuration shall provide a consistency rationale to ensure the compatibility of the combination of components (components are Base PPs and the PP-Modules that compose the PP-Configuration).
The consistency rationale shall demonstrate that the TOE overview is consistent with the TOE overview
of the PP-Configuration components and that the unions of SPDs, objectives, and security functional requirements defined in these components do not lead to a contradiction.
The consistency rationale may use correspondence tables between SPD/objectives/SFRs together with textual justifications._

woodbe:
Noted that this should be done, probably after the ICCC, as it seems the right thing to work towards.

@woodbe will start a PR for this when he has a chance.

Issue #188, opened 2019-10-18
Hybrid requirement placeholder


woodbe:
This is to hold the requirement FIA_HYB_EXT.1. At the Singapore meeting @lewyble stated that this requirement would be brought into MDFPP v3.3 as part of the integration since it really makes more sense there than in the PP-Module. 

As part of the "clean-up" to make the PP-Module more generic, this was taken out.

----------

=== MDFPP Security Functional Requirements Direction

This section instructs the ST author on what selections must be made to certain SFRs contained in the <<MDFPP>> in order to mitigate a threat from the <<MDFPP>> in a more specific or restrictive manner as described in this PP-Module than specified in the <<MDFPP>>.

*FIA_UAU.5 Multiple Authentication Mechanisms*

There is no change to the text of this SFR. However; the ST author must select at least one modality in FIA_UAU.5.1. The ST author shall select the same modality in FIA_MBV_EXT.1.1 in this PP-Module.

=== Identification and Authentication (FIA)

The following SFR shall be used by the ST author if 'hybrid' is selected in FIA_UAU.5.1.

==== FIA_HYB_EXT.1 Hybrid Authentication Biometric Method [[FIA_HYB_EXT.1]]

*FIA_HYB_EXT.1.1* The TOE shall only use [*selection*: _fingerprint, eye, face, voice, vein_], [*assignment:* _other modality_] as the biometric component of the hybrid authentication mechanism.

*Application Note {counter:remark_count}*:: A hybrid authentication mechanism is one where a user has to submit a combination of biometric sample and PIN or password with both to pass and without the user being made aware of which factor failed, if either fails. If this mechanism is selected in the <<MDFPP>>, the above component shall also be selected.

==== Hybrid Authentication Biometric Method (FIA_HYB_EXT)
===== Family Behaviour
This component defines the requirements for the TSF to be able to verify a user with the hybrid authentication.

===== Component leveling
[#img-FIA-HYB-EXT] 
.Component levelling 
[ditaa,"FIA_HYB_EXT.png"]
....
    +-----------------------------------------------------+
    |                                                     |     +---+
    | FIA_HYB_EXT  Hybrid authentication biometric method +---->| 1 |
    |                                                     |     +---+
    +-----------------------------------------------------+
....
 
FIA_HYB_EXT.1 Hybrid Authentication Biometric Method requires the TSF to verify a user with the hybrid authentication.

===== Management: FIA_HYB_EXT.1
There are no management activities foreseen.

[...]

PR #189, opened 2019-10-18
Move MDFPP references to PPC

woodbe:
The purpose of this set of changes is to remove the MDFPP references from the PP-Module and place them all into the PP-Configuration for the MDFPP.

During this edit I also changed the ECD images to be text-based (so they can be read directly). I also cleaned up the tables in the PPC.

I also created #188 to hold the FIA_HYB_EXT.1 requirement that should be moved into the MDFPP v3.3.


woodbe:
This is to close #187

woodbe:
The last 3 pull requests make the edits as discussed on the 11/21 call about removing "mobile" from the biometric enrollment/verification. This edited several SFRs as well as the title. I just removed the word mobile from them and left it to biometric in all cases.

woodbe:
Merged based on agreement on 12/5 call

PR #190, opened 2019-12-02
Move pad req to optional

n-kai:
This PR will update PP-Module, PP-Configuration and SD in Move-MDFPP-to-PPC branch to implement following changes;

1) Change PAD (FIA_MBV_EXT.3) to optional requirement in all documents
2) Change the format of the PP-Configuration referring the CC and latest draft of ISO/IEC 15408-1
3) Editorial changes (e.g. change "mobile biometric verification" to "biometric verification") in all documents

To implement above changes, you have to merge this PR into Move-MDFPP-to-PPC branch first and then merge #189 Move MDFPP references to PPC into master

### DIFF
--- a/3_ProtectionProfile/BiocPP.adoc
+++ b/3_ProtectionProfile/BiocPP.adoc
@@ -146,7 +146,7 @@ Zero-effort Impostor Attempt::
 - PP-Module Date: {revdate}
 
 === Base PP identification
-Base PP of this PP-Module is <<MDFPP>>.
+Base PP of this PP-Module is identified in the appropriate PP-Configuration.
 
 === TOE Overview
 ==== TOE main security features
@@ -217,7 +217,7 @@ The requirements for the TOE focus on the biometric performance (FTE, FAR/FMR an
 
 == Consistency rationale
 
-Consistency between the <<MDFPP>> and this PP-Module is demonstrated in the <<PPC-MDF>>.
+Consistency between the base PP and this PP-Module is demonstrated in the appropriate PP-Configuration.
 
 == Conformance Claims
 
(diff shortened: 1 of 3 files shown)


n-kai:
Sorry I can't understand your point. Could you tell me more about it or show me specific text you commented?

3_ProtectionProfile/BiocPP.adoc
```diff
@@ -129,16 +132,24 @@ Zero-effort Impostor Attempt::
...
+=== PP-Module Reference
+- PP-Module Reference: {doctitle}
+- PP-Module Version: {revnumber}
+- PP-Module Date: {revdate}
 
-=== PP Reference Identification
-- PP Reference: {doctitle}
-- PP Version: {revnumber}
-- PP Date: {revdate}
+=== Base PP identification
```
woodbe: Base-PPs should be defined in the PPC, not the PP-Module I would think, so this is one of the places where I have a problem with the MDFPP listing since it would imply I also need to list GPOS here as well.

3_ProtectionProfile/BiocPP.adoc
```diff
@@ -186,36 +193,50 @@ The evaluation of the above security functions is out of scope of this PP-Module
...
 
-This use case is an example of another use case that isn’t considered in this PP-Module. Another PP-Module should be developed at higher assurance level for this use case.
+This use case is an example of another use case that isn’t considered in this PP-Module. Another PP or PP-Module should be developed at higher assurance level for this use case.
 
 Mobile devices may be used for security sensitive services such as payment transactions and online banking. Verification may be done by the biometric for convenience instead of PIN/password to access such security sensitive services.
 
-The requirements for the TOE focus on the biometric performance (FTE, FAR/FMR and FRR/FNMR) and higher level of presentation attack.
+The requirements for the TOE focus on the biometric performance (FTE, FAR/FMR and FRR/FNMR) and presentation attack detection.
+
+== Consistency rationale
```
woodbe: I would expect this to be specified in the PPC since consistency would be part of the rationale of the PPC, not the PPM

3_ProtectionProfile/BiocPP.adoc
```diff
@@ -410,57 +410,47 @@ For consistency in language throughout this document, referring to a “lower”
...
-
-*Application Note {counter:remark_count}*:: Artificial PAIs that the TOE shall prevent and relevant criteria for its security relevant error rates for each type of PAI is defined in <<SD>>.
-
 === Protection of the TSF (FPT)
 ==== FPT_BDP_EXT.1 Biometric data processing [[FPT_BDP_EXT.1]]
 
 *FPT_BDP_EXT.1.1* The TSF shall process any plaintext biometric data used to generate templates and perform sample matching within the security boundary of the secure execution environment.
 
-*Application Note {counter:remark_count}*:: The Consistency Rationale in the appropriate PP-Configuration explains how the TOE in cooperation with its environment shall protect biometric data in detail.
+*Application Note {counter:remark_count}*:: The Consistency Rationale in the <<PPC-MDF>> explains how the biometric system in cooperation with the mobile device shall protect biometric data in detail.
```
woodbe: Here is the main place I'm talking about having specific references. The original states "appropriate PP-Configuration" while the update specifically points to the "PPC-MDF" configuration. This would mean the GPOS would not be able to match up without a new PPM update to add it here, and was the point of changing the text to be more generic and point to the appropriate PPC.

woodbe:
I am realizing in the SD there are a lot of places that should have MDFPP removed as well. For example, it is stated that crypto of the SEE should have been handled as part of the MDFPP evaluation, but this should probably state something along the lines of "crypto should be handled as part of the Base-PP".

This may be a little larger of a set of changes (and I'm not completely sure how I missed them originally, but obviously I did).

3_ProtectionProfile/BiocPP.adoc
n-kai: The reason why I add this "Base PP identification" is that CC Part1 requires PP-Module to include this identification. Please see "Figure 11 - PP-Module content" in CC part 1. Logically you are right but it may be better to follow CC.

*) https://www.commoncriteriaportal.org/files/ccfiles/CCPART1V3.1R5.pdf

3_ProtectionProfile/BiocPP.adoc
n-kai: I added this consistency rationale because the CC part 1 requires it (See figure 11 in CC Part1). I personally think that the CC should be modified and this rationale should be described in the PPC but we should follow the current CC.

3_ProtectionProfile/BiocPP.adoc
n-kai: OK, I didn't know your intention. I will update the text as you suggested. but how do you think about base PP identification/consistency rationale in the PP-Module?

n-kai:
OK now I understand your intention. I will update the SD to replace MDFPP with base PP or appropriate PPC so that PP-Module can be referenced by both GPOS and MDFPP at the same time.

3_ProtectionProfile/BiocPP.adoc
woodbe: I would think we could put in something here that would say that the consistency rationale is provided in an appropriate PPC, so we are referencing where to find it from here without needing to specifically place it or any references to the base-PP in this document. Maybe we also consider providing a reference to a web page (i.e. something on the github.io site) that lists the specific base-PPs the PPM has been approved to be used with so there is a clear reference to the other docs.

3_ProtectionProfile/BiocPP.adoc
woodbe: See my next comment about referencing

n-kai:
I removed references to MDFPP in SD and moved evaluation activities for SAR specific to MDFPP from SD to PPC.

woodbe:
Merged during 12/5 call

PR #198, opened 2019-12-06
Clarification for app note

### LINKED ISSUE #194: Application note 13
The wording of application note 13 seems like its adding a requirement to this SFR. My understanding is that application notes should add context/guidance to the SFR but are not required to be adhered too. The SD for the SFR states that this is out of the scope for the EA.

woodbe:
This is to clarify the app note. Several changes were made to try to be consistent with alternative auth factor, not just password auth factor.

This is to close #194.


woodbe:
I have added changes to the PPC and SD related to the alternative authentication factor for review.

PR #199, opened 2019-12-17
Update BiocPP.adoc

AlbertM-S:
I observe that "Mobile Device" is used as example of TOE Environment. As Mobile device have their own definition as “such as smartphones, PC login at offices, ATMs at banks, and building or room entrance control, or border security checks.” in the ESR. Probably could be interesting to change it as “i.e. smartphone”. Apart of that, I believe that the first TOE reference is in deed TOE Environment. 
What did you think?

Finally there are something that I do not change but probably will be interesting to check. Is the "Mobile Device" Glossary definition. Actually appear as “A self-contained device which is composed of a hardware platform and its system software. The device is typically some sort of general purpose computing platform, such as a laptop, tablet or smartphone that is designed to be portable.” Anyway in the ESR “Use Case(s)” explain “Biometric products are used for user authentication for mobile devices such as smartphones, PC login at offices, ATMs at banks, and building or room entrance control, or border security checks.”
Could have sense to change it as “A self-contained device which is composed of a hardware platform and its system software. The device is typically some sort of general purpose computing platform, such as smartphones, PC login at offices, ATMs at banks, and building or room entrance control, or border security checks.”

### DIFF
--- a/3_ProtectionProfile/BiocPP.adoc
+++ b/3_ProtectionProfile/BiocPP.adoc
@@ -150,7 +150,7 @@ Base PP of this PP-Module is identified in the appropriate PP-Configuration.
 
 === TOE Overview
 ==== TOE main security features
-This is a collaborative Protection Profile Module (PP-Module) used to extend a base PP for a mobile device that implement biometric enrolment and verification to unlock the mobile device in the locked state using the user’s biometric characteristics. Therefore, the Target of Evaluation (TOE) in this PP-Module is a mobile device that implements biometric enrolment and verification functionality. However, the term TOE in this document expresses the biometric system that is a part of the TOE (i.e. mobile device) and implements the biometric enrolment and verification functionality for clearly describing the relation and boundary between the biometric system and mobile device. Each biometric enrolment and verification process is described in the following paragraphs. 
+This is a collaborative Protection Profile Module (PP-Module) used to extend a base PP for a mobile device that implement biometric enrolment and verification to unlock the mobile device in the locked state using the user’s biometric characteristics. Therefore, the Target of Evaluation (TOE) in this PP-Module is a mobile device that implements biometric enrolment and verification functionality. However, the term TOE in this document expresses the biometric system that is a part of the TOE environment (i.e. smartphones) and implements the biometric enrolment and verification functionality for clearly describing the relation and boundary between the biometric system and mobile device. Each biometric enrolment and verification process is described in the following paragraphs. 
 
 a)	Biometric enrolment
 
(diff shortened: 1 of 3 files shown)


woodbe:
I agree with the idea, but I don't think using smartphone is the right way to go. I know the ESR may not have the "mobile" thing defined the way we want it to be here, but I think we need to define it some other way.

This is why I specifically defined "mobile device" here to be broader than the traditional version. Can we think of another term to be used instead? "Computing device"?

AlbertM-S:
I believe that computing Device Could work

woodbe:
Ok, so thinking about this some more, and a few points:

1. The ESR is actually more broad than what we are doing in this PP-Module. I expect at some point that the group will turn towards border control (an original topic) and it should be covered by the same ESR, but have much higher requirements. So to that end, the examples there are not necessarily the immediate concern, and I think we can be more broad in how we define things.
2. I plan to create a PP-Config that would work with GPOS as well (this would obviously need work with that TC, but I plan to update our initial draft once we are more "complete"), so I don't want to specify things like smartphones since we would be looking at other possible devices as well.

To that end though, I'm thinking "computer" is probably better than "computing device" since that is what we are talking about. A smartphone is a computer, just as is a laptop of a desktop, and in the definition I can be clear about the broadness (and limits) of this. I just think "computing device" is a weird term and technically can encompass a MUCH wider set of devices than we want, but computer is more accurate. 

Honestly I had left it a "mobile device" because I was a little lazy in that changing the definition to include traditional computers let me not have to replace it everywhere. I think changing it to computer with a similar definition to what is there now is probably best.

I will make this update if there is agreement.

n-kai:
@woodbe 
I agree to use "computer" as you proposed.

ccolin318:
I agree, I think computer will work.

woodbe:
I have made the change in the latest update.

woodbe:
@woodbe will check to remove smartphone from applicable places, will also computer definition, and verify any resulting changes that need to be made in the SD and PP-config

woodbe:
A few notes about the changes to the documents:

- The PP-Config only has a minor change to note at the beginning that a "mobile device" is called a "computer" in the PP-Module. I left this as mobile device here because this is how the PP-Module interacts with the MDFPP, which is all about mobile devices (i.e. not a "computer" in the broad sense).
- There are a few places where I did not remove PAI, or even added one. This was to specify (or try to) how an artefact relates to a PAI. I also specifically added a section in the SD to state that natural PAI were not covered and so when looking for new species, only artificial ones should be considered. I think this will work, since it would probably look a little odd if there was nothing about PAI in the documents at all.

PR #205, opened 2020-01-27
Correct title for SD in references

woodbe:
Add the full title for the SD in the document

Issue 3 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #206, opened 2020-01-27
tweak to artefact definition

woodbe:
Requested change was to change "specifically" to "specific", but I didn't think this quite worked as intended, so I adjusted the sentence a little more and placed "Specifically" at the beginning of the sentence.

Issue 4 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #207, opened 2020-01-27
addition of biometric characteristic definition

woodbe:
I also added "biometric" to the FNMR definition

Issue 5 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #208, opened 2020-01-27
add definition of Presentation

woodbe:
Issue 6 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #213, opened 2020-01-27
sync for O.Protection, OE.Protection

woodbe:
O.Protection, OE.Protection and OSP.Protection were not quite in sync about their claims of what was being protected.

Issue 15 from Cybersecurity Malaysia


woodbe:
I have made these adjustments (I was just making their changes, not fully thinking about how they would play out at the time).

woodbe:
Merged based on 2/13 call.

PR #216, opened 2020-01-27
Adjust descriptions of non-mandatory requirements

woodbe:
The statements in the Selection base requirements is confusing as it talks about all the types that aren't mandatory.

They should be split out into the different sections instead.


woodbe:
> Line 465 "The this section" should be changed to "This section"

This has been fixed.

woodbe:
Merged based on 1/30 call

PR #217, opened 2020-01-27
enrollment

woodbe:
misspelling of enrolment (found in 2 places)

Issue 19 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #218, opened 2020-01-27
PA requirement descriptions should mention detection

woodbe:
FIA_MBE_EXT.3 should mention that PA should be detected as well as prevented.

@woodbe also updated FIA_MBV_EXT.3 since it is the same for verification.

Issue 20 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #219, opened 2020-01-27
Diometric misspelling

woodbe:
Biometric misspelled in FPT_BDP_EXT component leveling diagram

Issue 21 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #227, opened 2020-01-27
Change utility to quality

woodbe:
Reads better and seems more accurate

Issue SD3 from Cybersecurity Malaysia


woodbe:
Merged based on 1/30 call

PR #256, opened 2020-02-10
Editorial fixes based on NK public review comments

n-kai:
Updated based on my public review comments modified by Brian


woodbe:
Merged based on 2.13 call

PR #230, opened 2020-01-27
add test step to FIA_MBV_EXT.2

woodbe:
Add another item to check

Issue SD6 from Cybersecurity Malaysia

PR #259, opened 2020-02-14
update to add assessment criteria link

woodbe:
Added a link for all parts where the assessment criteria for templates is mentioned, so it points back to the original location in the document where it was mentioned.

This is to close #230


woodbe:
Accidentally made a new PR instead of adding to existing PR.

PR #266, opened 2020-03-05
Editorial updates for proposed release

### LINKED ISSUE #255: review all internal links for consistency
some things that should be linked were not ([CC2], [CC3], maybe more), some things that were linked didn't have [].

woodbe:
This is a set of editorial updates in preparation for the proposed release. All the docs have been set to a version of 0.95 ad a date of March 13.

Note that I changed "test document" in the BIOSD to "performance report" and linked it everywhere to that section except within that section (since that didn't seem to make sense).

This will close #255


4_Methodology/BS_SD.adoc
```diff
@@ -757,46 +766,48 @@ The evaluator shall report the summary of result of EA defined above, especially
...
+To solve this issue, the Biometric Security iTC (BIO-iTC) created and maintains the PAD <<Toolbox>>. <<Toolbox>> defines the common artefacts for PAD testing based on publicly available information (e.g. research papers), experiences and knowledge shared among the BIO-iTC members.
 
 <<Toolbox>> includes a collection of test items for each biometric modality. Each test item describes the procedure to create artefacts and the method to present them to the TOE in sufficient detail to enable the test to be repeatable.
 
 The same <<Toolbox>> can also be used for AVA_VAN.1 evaluation (i.e. penetration testing) because AVA_VAN.1 requires the evaluator to devise tests based on information available in the public domain. However, <<Toolbox>> should be used in a different manner for AVA_VAN.1 evaluation. The following section explains how <<Toolbox>> should be used in EAs for ATE_IND.1 and AVA_VAN.1.
 
+<<Toolbox>> is intended to be a state of the art set of test and is continually updated. The latest versions of the <<Toolbox>> for each supported modality must be used by the evaluator at the time the testing begins.
```
woodbe: I am not sure this should be here or not. It seemed like it was needed. This could also just be moved (or edited).

4_Methodology/BS_SD.adoc
```diff
@@ -757,46 +766,48 @@ The evaluator shall report the summary of result of EA defined above, especially
...
 
 During the independent testing, the evaluator may find artefacts that are incorrectly matched to the enrolled target user however, the evaluator may not be able to reliably reproduce a successful presentation attack.
 
-<<Toolbox>> defines the Pass/Fail criteria, maximum imposter attack presentation match rate for artefacts. The evaluator shall follow the <<Toolbox>> criteria for the number of artefact presentations and confirm that the TOE’s match rate is below the specified criteria during the independent testing. The evaluator shall assign a fail verdict to those TOE that doesn’t satisfy the criteria.
+<<Toolbox>> defines the Pass/Fail criteria, maximum imposter attack presentation match rate for artefacts. The evaluator shall follow the <<Toolbox>> criteria for the number of artefact presentations and confirm that the TOE’s match rate is below the specified criteria during the independent testing. The evaluator shall assign a fail verdict to those TOE that don’t satisfy the criteria.
 
 The artefacts that pass the criteria but show the higher imposter attack presentation match rate will be tested again during the AVA_VAN.1 evaluation.
```
woodbe: Is it correct that this is the case? If new artefacts are found it doesn't seem to require using the existing toolbox tests to check, even if there aren't many new (i.e. say only one new artefact is found).

If this really is a "will be tested" then we should adjust the section about what to test to include the highest likelihood artefacts here as well as anything new.

woodbe:
I have also updated the generic relation diagram in the PPM as it still said mobile device instead of computer. Both the PNG and the PPTX have been uploaded as part of this.

4_Methodology/BS_SD.adoc
n-kai: Is this a comment for line 801? I am not sure that my comment below correctly deal with your concern but see below.

I think that we should update the toolbox as follows.

1. If the evaluator can find a new artefact that are not included in the toolbox during the penetration testing, the evaluator shall inform the BIO-iTC (See 2nd para in 6.3.1.3.2. New artefacts found test plan [BIOSD])
2. the BIO-iTC verify that the new artefact is actually within the Basic attack potential and add it to the toolbox 
3. the BIO-iTC informs the toolbox update to CCRA

woodbe: Sorry, the comment didn't link very well. Yes, this is about line 801. If you also look at the section starting at 

https://github.com/biometricITC/cPP-biometrics/blame/eeb8a1ff3903f88a4eb55e59f6c9bab1d81810ca/4_Methodology/BS_SD.adoc#L911

the test plan for new artefacts found implies that you then ONLY test the new artefacts. There isn't anything that states that if you find there to be new artefacts that you should test those in addition to the most likely candidates from the IND testing, which I think is what we actually want. 

So what I'm really getting to is that if you can find new artefacts to test, and we only look at those, then saying that this potential candidate "will be tested" is incorrect since it wouldn't be covered under the test plan. 

The way we have the test plans laid out, there is one set of steps to follow if you have new artefacts on one if you don't. I think that what we really want is one test plan that ALWAYS covers these likely candidates, and IF new artefacts are found, also tests those. Today though, I read it as two independent test plans/paths.

So if we MUST always test these high potential candidates from the IND, then we need to adjust how the AVA_VAN test plan is written out so ensure both cases are tested in all scenarios. Or we need to say that these high potential candidates MAY be tested.

n-kai: @woodbe 
I created PR #269 to update your editorial-updates branch for the above issue. Please take a look at it and tell me if it can't work.

4_Methodology/BS_SD.adoc
n-kai: I suggest to move the above text (Line 797) to 2. Biometric modalities supported in Toolbox Overview (https://github.com/biometricITC/cPP-toolboxes/blob/develop/toolbox.adoc#biometric-modalities-supported)

woodbe: I have moved the line to the Toolbox.

woodbe:
Agreed to merged on 3/11 call

PR #269, opened 2020-03-09
patch for new artefacts

n-kai:
I propose some text to address issue related to new artefacts testing (See #266)

### DIFF
--- a/4_Methodology/BS_SD.adoc
+++ b/4_Methodology/BS_SD.adoc
@@ -882,7 +882,7 @@ The evaluator shall perform EAs in <<No new artefacts found test plan>> if there
 
 The evaluator shall select those artefacts that show higher imposter attack presentation match rate at the independent testing. The evaluator shall test them extensively during the penetration testing.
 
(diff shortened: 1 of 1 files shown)


woodbe:
This still doesn't address my concern, though I do think it is a good update. I still see the testing as this (some psuedo code to maybe help):

Are there NEW artefact types?

YES -> Are they within the attack potential?

  **YES 2 -> Run tests with new artefacts -> DONE**

  NO 2 -> Proceed as with no new artefacts -> NO

NO -> Take likely candidates from ATE_IND and generate "higher quality" artefacts to test with

END

Note that at the YES 2 phase, the lab would test only the new artefacts, and not any of the likely candidates from the ATE_IND testing. So if you say yes there are new artefacts, and they are within the attack potential, the current docs read that those are all you test (in how I read them). 

I think this is what we want:

  YES2 -> Run tests with new artefacts & Take likely candidates from ATE_IND and generate "higher quality" artefacts to test with

n-kai:
I thought that I updated text following the last YES2 but it seems my update didn't work. I updated my patch branch again.

PR #270, opened 2020-03-20
Update PP_Config.adoc

woodbe:
Found a few mismatches to the MDF and removed some links in a section heading. I was looking at the GPOS and noticed these while comparing them together.

These should get merged as part of the final update.


woodbe:
Merged based on agreement on 3/26 call

PR #271, opened 2020-03-20
Create PP_Config-OS.adoc

woodbe:
Initial updated GPOS PP-Config.

This is based on the latest MDF version.

The main concern is section 5.4.1.5 which talks about the SEE. There isn't an SEE or equivalent required for the OS (which makes sense as it is not hardware). This could mean we can't make this work without more than some basic edits to the PP (i.e. auth edits to allow biometrics).

One possible method though would be to only allow approval based on some expectation that the biometric subsystem is isolated hardware and does not use the AP of the computer. So a self-contained biometric system using its own processor (I may say it could allow storage on the disk if it encrypts it first, like a TEE), would be the only acceptable devices that could be used here. A USB (or maybe BT) device would work, but not one where it handed everything off to the OS to process and store.

I'm not sure this will work, it may be layering too many requirements on how this should work.

I'm sending this version to Mike Grimm at Microsoft (since they are the company that validates end user OS, I don't know that Oracle would care about this topic).

### DIFF
--- a/3_ProtectionProfile/PP_Config-OS.adoc
+++ b/3_ProtectionProfile/PP_Config-OS.adoc
@@ -16,7 +16,7 @@ This PP-Configuration was developed by the Biometrics Security international Tec
 
 The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that combines Protection Profiles (PPs) and PP-Modules for various technology types into a single configuration that can be evaluated as a whole. The scope includes the definition of the configuration of a computer (a computer in the terms of the PP-Module) that has biometric enrolment and verification capability. The TOE will be defined by a combination of the components described in <<PP-Configuration Components Statement>>.
 
-The computer for the purposes of this PP-Configuration is a system with a General Purpose Operating System plus a Biometric System. The Biometric System may be directly integrated into the physical computer or an external component that has been linked/paired with the physical computer and for which biometric credentials are linked specifically to the paired computer (i.e. templates registered on computer 1 cannot be used on computer 2).
+The TOE for the purposes of this PP-Configuration is a system with a General Purpose Operating System plus a Biometric System. The Biometric System may be directly integrated into the TOE, or may include an external component that has been incorporated as part of the TOE platform and for which biometric credentials are linked specifically to the TOE platform (i.e. templates registered on TOE platform 1 cannot be used on TOE platform 2).
 
 == PP-Configuration Reference
 
@@ -41,7 +41,7 @@ This section describes consistency rationale between <<PP_OS_V4.3>> and <<BIOPP-
 
 === Consistency of TOE Type
 
-When the <<BIOPP-Module>> is used to extend <<PP_OS_V4.3>>, the TOE type for the overall TOE is still a general purpose operating system. However, one of the functions of the operating system and computer must be the ability for it to have biometric enrolment and verification capability. The TOE boundary is simply extended to include that functionality.
+When the <<BIOPP-Module>> is used to extend <<PP_OS_V4.3>>, the TOE type for the overall TOE is still a general purpose operating system. However, one of the functions of the TOE must be the ability for it to have biometric enrolment and verification capability. The TOE boundary is simply extended to include that functionality.
 
 === Consistency of Security Problem Definition
 
@@ -71,7 +71,7 @@ The threats, OSPs and assumptions defined by the <<BIOPP-Module>> (see the Secti
 |Consistency Rationale
 
 |A.Alternative	
-.3+|All assumptions levied on the operational environment of biometric system (i.e. The computer) are consistent with security requirements in the <<PP_OS_V4.3>>. 
+.3+|All assumptions levied on the operational environment of biometric system (i.e. the computer) are consistent with security requirements in the <<PP_OS_V4.3>>. 
 |A.Authentication
 |A.User
 
@@ -103,7 +103,7 @@ The objectives for the biometric system and its operational environment are cons
 |Consistency Rationale
 
 |OE.Alternative	
-.4+|All Environmental Objectives levied on the operational environment of biometric system (i.e. The computer) are consistent with security requirements in the <<PP_OS_V4.3>>. 
+.4+|All Environmental Objectives levied on the operational environment of biometric system (i.e. the computer) are consistent with security requirements in the <<PP_OS_V4.3>>. 
 |OE.Authentication
 |OE.Protection
 |OE.User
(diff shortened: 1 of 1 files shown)


woodbe:
To be clear to everyone, this is not a priority to finish for our release, but I wanted to see if I could get some feedback during our last review that could point out adjustments to make that would allow easier integration with GPOS in the future. I am not sure anything will come in that timeframe though.

n-kai:
@woodbe I understand your concern about creating SEE in the OS. so we may have to add additional assumptions for the OS or modify the PP as you mentioned. But I need more time to think about it. But your first version looks pretty good.

3_ProtectionProfile/PP_Config-OS.adoc
```diff
@@ -0,0 +1,333 @@
...
+
+:sectnums:
+:sectnumlevels: 5
+
+== Introduction
+
+The purpose of a PP-Configuration is to define a Target of Evaluation (TOE) that combines Protection Profiles (PPs) and PP-Modules for various technology types into a single configuration that can be evaluated as a whole. The scope includes the definition of the configuration of a computer (a computer in the terms of the PP-Module) that has biometric enrolment and verification capability. The TOE will be defined by a combination of the components described in <<PP-Configuration Components Statement>>.
+
+The computer for the purposes of this PP-Configuration is a system with a General Purpose Operating System plus a Biometric System. The Biometric System may be directly integrated into the physical computer or an external component that has been linked/paired with the physical computer and for which biometric credentials are linked specifically to the paired computer (i.e. templates registered on computer 1 cannot be used on computer 2).
```
The-Fiona: "... may be directly integrated into the physical computer or an external component that has been linked/paired with the physical computer..." seems to go against what the GPOS PP defines as the TOE. The GPOS TOE includes physical and virtual hardware, so I guess the config should try and allow this too.  Here I Mae a suggestion, but it may be less than perfect...

Suggest "The Biometric System may be directly integrated into the TOE platform, which may include an external component that has been incorporated as part of the TOE platform and for which biometric credentials are linked specifically to the TOE platform (i.e. templates registered on TOE platform 1 cannot be used on TOE platform 2).

3_ProtectionProfile/PP_Config-OS.adoc
```diff
@@ -0,0 +1,333 @@
...
+
+The following rationale identifies several SFRs from <<PP_OS_V4.3>> that are needed to support Biometric System functionality and explains why the unions of SFRs in the <<PP_OS_V4.3>> and <<BIOPP-Module>> do not lead to a contradiction.
+
+==== Relation among SFRs/OEs in the PP_OS_V4.3 and BIOPP-Module
+The relation between SFRs defined in the <<PP_OS_V4.3>> and SFRs and OEs in the <<BIOPP-Module>> is described below for each security functionality. *Bold SFRs* are those SFRs defined in the <<BIOPP-Module>> for the Biometric System and _italicized SFRs_ are those defined in <<PP_OS_V4.3>> for the computer.
+
+===== Authentication Factors
+The authentication factors defined in the <<PP_OS_V4.3>> are Non-Biometric Authentication Factors as defined in the <<BIOPP-Module>>. The computer shall implement an authentication factor as required by the _FIA_UAU.5.1._ These authentication factors are used as an alternative authentication mechanism when the user is rejected by the biometric verification.
```
The-Fiona: computer

Replace
"The _computer_ shall implement an authentication factor as required by the _FIA_UAU.5.1._ These authentication factors are used as an alternative authentication mechanism when the user is rejected by the biometric verification"
with:
"The **TOE** shall implement an authentication factor as required by the _FIA_UAU.5.1._ These authentication factors are used as an alternative authentication mechanism when the user is rejected by the biometric verification"

3_ProtectionProfile/PP_Config-OS.adoc
```diff
@@ -0,0 +1,333 @@
...
+
+The <<BIOPP-Module>> assumes that above requirements are satisfied by the computer as defined in OE.Alternative.
```
The-Fiona: Replace "_computer_" with "**TOE**"

3_ProtectionProfile/PP_Config-OS.adoc
```diff
@@ -0,0 +1,333 @@
...
+.4+|All Environmental Objectives levied on the operational environment of biometric system (i.e. The computer) are consistent with security requirements in the <<PP_OS_V4.3>>. 
+|OE.Authentication
+|OE.Protection
+|OE.User
+
+|===
+
+=== Consistency of Requirements
+
+The Biometric System (i.e. TSF in the <<BIOPP-Module>>) is comprised of biometric capture sensors and firmware/software that provide functions described in the <<BIOPP-Module>> TOE design. The Biometric System may be integral to the computer or an external (self-contained) device that has been linked/paired with the computer. The Biometric System is invoked by the computer as defined in the <<PP_OS_V4.3>> when user’s biometric characteristics is presented to the sensor. The Biometric System creates and stores the template or compares the features with the stored template and returns the verification outcome to the computer.
```
The-Fiona: 1.
Replace..."The Biometric System may be integral to the computer or an external (self-contained) device that has been linked/paired with the computer."...
with:
"The TOE platform includes a Biometric System that may be integral to the GPOS TOE platform or extended to include an external (self-contained) device that has been linked/paired with the GPOS TOE platform.

2. 
"The Biometric System is invoked by the computer as defined in the <<PP_OS_V4.3>>"
"The Biometric System is invoked by the TSF as defined in the <<PP_OS_V4.3>>

3.
Replace
"The Biometric System creates and stores the template or compares the features with the stored template and returns the verification outcome to the computer"
with:
"The Biometric System creates and stores the template or compares the features with the stored template and returns the verification outcome to the TOE."

3_ProtectionProfile/PP_Config-OS.adoc
```diff
@@ -0,0 +1,333 @@
...
+
+The <<BIOPP-Module>> assumes that the computer satisfies SFRs defined in the <<PP_OS_V4.3>> so that the Biometric System can work as specified in the <<BIOPP-Module>>. This section explains which SFRs in the <<PP_OS_V4.3>> are directly relevant to the Biometric System security functionality.
```
The-Fiona: Replace:
"The <<BIOPP-Module>> assumes that the computer satisfies SFRs defined in the <<PP_OS_V4.3>> so that the Biometric System can work as specified in the <<BIOPP-Module>>."

with

"The <<BIOPP-Module>> assumes that the TOE satisfies SFRs defined in the <<PP_OS_V4.3>> so that the Biometric System can work as specified in the <<BIOPP-Module>>.

Note that in the case where the TOE platform is "extended" with additional biometric hardware/firmware/software the GPOS SFRs might/should extend to the new "larger TOE".

woodbe:
@The-Fiona based on your comments here, do you see any issues with the SEE and the GPOS evaluation?

woodbe:
I picked this back up today, and I think we will have to put this off until later. On looking at this further, here is what I see.

In the OS_PP, the hardware is specifically out of scope (from the TOE Platform):
> The TOE platform, which consists of the physical or virtual hardware on which the TOE executes, is outside the scope of evaluation. At the same time, the security of the TOE relies upon it. Other hardware components which independently run their own software and are relevant to overall system security are also outside the scope of evaluation.

Because we are expecting the SEE to be evaluated as part of the base PP, and that is out of scope, it would seem like we have to make several assumptions. I think these will be a step too far overall with how we have written the PP-Module right now, especially since I don't know when this would actually be able to be combined.

Looking closer at the MDF, while everyone is using some sort of SEE, technically outside of the root keys, it seems like you could probably technically get away without one on a mobile device as well (though I don't see that happening), so we may need to consider how we want to strengthen the SEE expectations in the future, which could help point to the way forward for the OS_PP integration.

From the OS_PP standpoint, what I think would be required, is to assume that the sensor subsystem is a discrete component that includes its own storage (or like a TEE, I guess it could encrypt the files and store them outside) and has its own processor for everything. We would then use something like O.PROTECTED_COMMS to say that the communications between the sensor system and the OS are protected in some manner (and maybe O.PROTECTED_COMMS needs to include something about hardware linkage, though I'm not sure how that happens given the TOE Platform statements). 

We could then write some requirements on the expected security of the sensor system (for example a USB sensor that links into the auth system, but handles all decisions internally and only returns YES/NO to the OS on an auth request). This could be handled by some descriptions, maybe in a PP-Module (so you could evaluate for OS_PP with the performance requirements we have now and the environment requirements), or by a combination with something like DSC (though that is likely to be very heavy to do).

This definitely isn't something we can handle fully in 2-4 weeks, I would expect a few months, and back and forth with the OS TC (which won't happen now with the NIAP shutdown), so I think this needs to be put on hold for a future update that would also incorporate some changes to the PP-Module as well. Maybe this is a v1.1 thing.

woodbe:
Merged after v1.0 published

PR #274, opened 2020-04-21
Update for attack potential

### LINKED ISSUE #272: Attack Potential table
Attack potential table in the BIOSD

https://github.com/biometricITC/cPP-biometrics/blame/master/4_Methodology/BS_SD.adoc#L1398

does not match the Table 4 in the CEM (page 428). 

Our table just goes in increments of 10, while the CEM has a different set of levels. I think we should be using the CEM as we did not create our own listings. It looks like only the values are wrong.

Also, the value calculations do not match up with the CEM.

I don't know where these values come from as I didn't see an explicit reference in the BIOSD for it, but it seems like these probably need to match the CEM documentation.

I realized this when I was looking at the eye toolbox and moving it over.

woodbe:
This is to close #272 with the update to the attack potential table based on the updated categories from @n-kai

In addition I cleaned up the tables some, there there are a lot of changes, but the changes are mainly just moving the rows into separate lines so it is a little easier to read in the document. I also made a few minor edits to the formatting while cleaning up the tables.

The main changes are in lines 1236, 1318-1320, 1324 and 1622 (there are some other places in the tables where I added in the new content, but these are the main new pieces I have added).


4_Methodology/BS_SD.adoc
```diff
@@ -1314,17 +1314,14 @@ However, it is important to be able to compare the resistance of various systems
...
-. _Difficult_ is for iris and vein. Iris images can be acquired with a high resolution camera, but with some difficulties to get a complete high quality image without the cooperation of an enrolled data subject. Veins are a hidden characteristic, but infra-red cameras, close to them, can acquire images to be used.      
---
+. _Without notice_ is for samples for making an artefact that can be collected without any contact with an enrolled subject. For example, 2D face images uploaded on the Internet and latent fingerprint images on a glass can be collected without notice of the subject.
+. _Non-cooperative_ is for samples for making an artefact that need to be collected directly from an enrolled subject in a short period of time without full cooperation from the subject. For example, iris or vein images need to be acquired with a high resolution or infra-red camera, however, such images can be taken in a moment without full control of the subject.
```
gfiumara, line 1319:
```suggestion
. _Non-cooperative_ is for samples for making an artefact that need to be collected directly from an enrolled subject in a short period of time without full cooperation from the subject. For example, iris or vein images need to be acquired with a high resolution or infrared camera, however, such images can be taken in a moment without full control of the subject.
```

4_Methodology/BS_SD.adoc
```diff
@@ -1314,17 +1314,14 @@ However, it is important to be able to compare the resistance of various systems
...
+. _Cooperative_ is for samples for making an artefact that need to be collected directly from an enrolled subject with full cooperation from the subject. For example, the acquisition of a detailed 3D face scan of the subject takes time and requires full cooperation from the subject.
```
gfiumara, line 1320:
```suggestion
. _Without notice_ is for making an artefact with samples that can be collected without any contact with an enrolled subject. For example, 2D face images uploaded on the Internet and latent fingerprint images on a glass can be collected without notice of the subject.
. _Non-cooperative_ is for making an artefact with samples that need to be collected directly from an enrolled subject in a short period of time without full cooperation from the subject. For example, iris or vein images need to be acquired with a high resolution or infra-red camera, however, such images can be taken in a moment without full control of the subject.
. _Cooperative_ is for making an artefact with samples that need to be collected directly from an enrolled subject with full cooperation from the subject. For example, the acquisition of a detailed 3D face scan of the subject takes time and requires full cooperation from the subject.
```

4_Methodology/BS_SD.adoc
```diff
@@ -1314,17 +1314,14 @@ However, it is important to be able to compare the resistance of various systems
...
 
 *Application Note {counter:remark_count}*:: The above distribution of modalities per level is subject to modification depending on the evolution of technologies and usage. The current distribution is to be seen as guidance for the evaluator, who will have to adapt the rating to state-of-the-art.
```
gfiumara, line 1322: This application note no longer applies, since you are not specifying modalities.
```suggestion
```

4_Methodology/BS_SD.adoc
woodbe: infrared seems to be the common spelling today, not infra-red. Even dictionary.com points infra-red to the infrared description and doesn't even try to show the work with the hyphen (outside some search result examples).

woodbe:
Merged as approved on 4/23 call

PR #285, opened 2020-05-05
Updates for public release

woodbe:
These are focused on updating the dates and versions for all the documents in preparation of public release on May 11, 2020.

Note that the PPC is listed as v0.99 instead of 1.0 because the MDFPP hasn't been updated yet, so this is to really just mark that it is done, but not finished since the MDFPP isn't ready yet.

The group can determine if the version should be 1.0 instead of .99.


woodbe:
This would be merged after all other pull requests/changes are approved.

Issue #295, opened 2020-02-27
Low sample quality


woodbe:
The overview for each toolbox should have a section that discusses the expectations regarding low quality samples. While the discussion may not be exhaustive, it should talk about common ways to create low quality samples and how that can be used. This is a corollary to biometricITC/cPP-biometrics#262 to be added into the BIOSD.

PR #298, opened 2020-10-23
Sample quality TD

woodbe:
Proposed changes for the TD to add the sample quality research paper and an explicit reference for the validator to point to this as a reference.


woodbe:
This is to close #295

gfiumara (approved):

n-kai (approved):

PR #311, opened 2021-03-04
scope specification

### LINKED ISSUE #301: [NIAP] Scope specification
Section 3.3.3, p. 9, before figure: This is really a general comment, but the way that TOEs will be evaluated against the requirements described in this PP-Module will be in the context of a PP-Configuration, which will include the requirements of the base PP. So, it's not correct to say "above security functions [are] out of scope", because during the TOE evaluation those requirements will be in scope for the TOE. A more correct statement would be that the *specification* of the functions is out of scope of the PP-Module, and are specified in the base PP.

woodbe:
This is to close #301

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -206,7 +206,7 @@ The TOE is reliant on the computer itself to provide overall security of the sys
 * Invoking the TOE to enrol and verify the user and take appropriate actions based on the decision of the TOE
 * Providing the Secure Execution Environment that guarantees the TOE and its data to be protected with respect to confidentiality and integrity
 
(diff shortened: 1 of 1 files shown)


gfiumara:
> and expected to be included

If we want to require this, should we say **shall** be included in the base PP?

woodbe:
@gfiumara how about replacing "expected to be included as" with "are"

so it is just a statement that they will be in the base PP (since if they weren't in the base PP we wouldn't have agreed to the PP-Configuration and support for the base PP).

gfiumara:
> @gfiumara how about replacing "expected to be included as" with "are"

Yes, I like that much more.

gfiumara (approved):

n-kai (approved):

Issue #314, opened 2021-03-30
[cPP CHANGE] Promote application note regarding sufficient quality


gfiumara:
**What is the change request for the cPP/PP-module? Please describe.**

Application note:
> ST author may refine "sufficient quality" to specify quality standards if the TOE follows such standard.

I am thinking this application note should be appended to FIA_MBE_EXT.2.1. I think it enhances the overly-ambiguous "sufficient quality" and avoids normative language in an application note (#304). 

**Describe the solution you'd like**
> *FIA_MBE_EXT.2.1* The TSF shall create templates of sufficient quality. If the TOE follows a specific quality standard, ST author shall quantify "sufficient quality."

This also promotes the MAY to a SHALL.

**Additional context**
See also #304.

woodbe:
I think we may have taken this from ISO work (I remember we took SFRs from that early on, I don't fully remember the provenance though). That said, we could adjust the SFR to be something more like:

> The TSF shall only use samples that meet [assignment: specify quality standards] to verify the user.

From there we could adjust the app note to clarify that the ST author could either point to a specific standard being used or to define their internal standard. Maybe this is a little less ambiguous (maybe)

woodbe:
The current requirement in the MDF reads as this:

> The TSF shall only use biometric samples of sufficient quality for enrolment. As such, sample data shall have [assignment: quality metrics corresponding to each biometric modality].

There is a similar requirement for verification (which we will need to change here also). 

I think the big thing here is that for the MDF, this is an objective requirement, so the vendor can select it as needed, while we make it mandatory, and so need to provide a means for specifying something. The issue, as I see it, is that a vendor may not actually meet some externally defined standard (such the NFIQ), but have their own. This means to me that we need to either guide the vendor via the app note that they can fill in anything in the assignment, or that we explicitly put a selection there that says:

[selection: no independent quality metric, [assignment: quality metric standard]]

I don't think we need to use the "corresponding to each biometric modality" here.

And then an "independent quality metric" would be defined as some sort of external (i.e. not defined by the vendor) standard, such as NFIQ, that could be referenced, while if the "no" option is specified, the vendor would have to provide a description of the metric that is used to determine quality in the TSS (confidential portion). And the SD should provide some idea as to what the evaluator should look for in such a document (if what we have currently isn't sufficient, I'm not certain).

Issue #304, opened 2021-03-02
[NIAP] Application notes not normative


woodbe:
Section 7.1, 7.2 (throughout): Application notes cannot be normative; therefore, there should be no "shall" statements in application notes. Further, the objectives are merely high-level statements of TOE functionality applicable for the corresponding threats/OSPs/assumptions; they are not requirements either. So, the use of "shall" in these statements is also not recommended.

Section 8 (throughout): The "shall" statements in application notes are particularly troubling here, since they appear as requirements to be addressed during the TOE evaluation. For example, application note10 is a timing of authentication requirement. If this is really a "shall" then the either another requirement needs to be included in the PP-Module, or an appropriate requirement in the Base PP needs to be identified/modified/etc. to provide this functionality.

woodbe:
I'm closing this as the remaining issue for this is in #314

PR #315, opened 2021-03-30
Remove normative language from application notes

gfiumara:
* Towards #304.
 * Leaves one application note with normative language and raises Issue #314.
 * One grammar change.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -413,8 +413,6 @@ This section lists SFRs for the biometric enrolment and verification.
 
 *FIA_MBV_EXT.1.2*:: The TSF shall provide a biometric verification mechanism with the [*selection*: _FMR, FAR_] not exceeding [*assignment*: _defined value_] and [*selection*: _FNMR, FRR_] not exceeding [*assignment*: _defined value_].
 
-*Application Note {counter:remark_count}*:: If the TOE supports multiple modalities, not all modalities need to define the Evaluation Activities.
-
 *Application Note {counter:remark_count}*:: Consider the following factors when setting values of FMR, FAR, FNMR and FRR.
 +
 [loweralpha]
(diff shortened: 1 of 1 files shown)


Protection Profile/BiocPP.adoc
```diff
@@ -413,16 +413,16 @@ This section lists SFRs for the biometric enrolment and verification.
 
 *FIA_MBV_EXT.1.2*:: The TSF shall provide a biometric verification mechanism with the [*selection*: _FMR, FAR_] not exceeding [*assignment*: _defined value_] and [*selection*: _FNMR, FRR_] not exceeding [*assignment*: _defined value_].
 
-*Application Note {counter:remark_count}*:: If the TOE supports multiple modalities, the ST author may iterate the SFR to define different error rates for each modality.
+*Application Note {counter:remark_count}*:: If the TOE supports multiple modalities, the error rates do not to be identical.
```
woodbe, line 416: I'm not sure if we need to specify whether this requirement can be iterated or not. That is the norm for having different answers (since you can define the assignment as to a specific option), but I'm not sure if that convention is needed or not.

@n-kai what do you think?

n-kai: @woodbe 
I agree with you. It's better to remove the last application note (I checked MDF PP and related PP modules but no one has such application note)

gfiumara: I have removed this application note in a39029230239343e07bcf98b9a5b488e6c88cb92. Should we also remove the next note that seems to also correspond to iteration?

> If the TOE supports multiple modalities, not all modalities need to define the Evaluation Activities.

woodbe: OK, I'll remove it

woodbe: @gfiumara yes, I would remove that one too

woodbe:
I removed the second application note with https://github.com/biometricITC/cPP-biometrics/pull/315/commits/2fa17d795af5f7f5e440a85fc504f904086de4cf

woodbe (approved):

Issue #310, opened 2021-03-02
[NIAP] SD strict guidance needed


woodbe:
SD (general): There are several items listed (assessment criteria for samples, for example) that are provided by the vendor but used as SD-like guidance for the evaluators. Without strict guidance on what these have to contain, the vendor can write anything. This does not to fit with the objectives of repeatable, consistent evaluation activities for the same requirements.

woodbe:
@ccolin318 while there are several clear sections around the assessment criteria, do you have any other specific areas where you think we should look hard (I think I saw a few minor ones, but right off the top of my head I don't recall them).

woodbe:
I haven't started writing this section yet, but I am looking at the "Developer’s performance report and its assessment strategy" section as the basis for it. I don't think we will need to get into individual details at this point for each modality, but we don't have an outline of what we expect listed anywhere. 

My thought really is that we take the Biometric quality paper we have linked and put together an expectation that if they aren't pointing to something like NFIQ or ISO 29794-4/5/6 (at least for face/fingerprint/iris, I haven't been able to find anything for standards like this for vein, though I did find #341), that they need to provide some sort of analysis that would be based on one of these. So I think that we can write something up using the information in the quality reviews to provide a guide for the assessment in general (and the evaluator already should be familiar with these papers for more information about the modalities), and we will be good.

Does anyone have any thoughts about this approach? I will try to complete it by the end of next week.

n-kai:
@woodbe 
What kind of evaluation activities should be defined in the BIOSD? Are you thinking of the following activities?
- The evaluator needs to review the BMD to check that TOE's quality scoring algorithm conforms to those standard or technical paper. (However, TOE's quality scoring algorithm may be customized and evaluator may see some deviation from those standard or paper)
- The evaluator needs to test the TOE that the TOE reject any samples that are below the quality score

n-kai:
@woodbe 
_Does anyone have any thoughts about this approach? I will try to complete it by the end of next week._
I agree with your direction and don't have any other questions so please go ahead and create the PR.

PR #318, opened 2021-04-16
Quality metric update

### LINKED ISSUE #314: [cPP CHANGE] Promote application note regarding sufficient quality

woodbe:
This is to close #314

This is a modification from the MDF v3.2 requirement to make it explicit that a vendor can choose their own metric (with a description to be specified) or an independent one (such as NFIQ, though I haven't added a reference for it, and I'm not sure if we should in the app note or not).

I added some bullets in the SD to note this as well.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -430,7 +430,7 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_MBV_EXT.2 Quality of biometric samples for biometric verification [[FIA_MBV_EXT.2]]
 
-*FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _no independent quality metric, [*assignment*: quality metric standard]_].
+*FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _an internal quality metric_, [*assignment*: _quality metric standard_]], using a threshold of [*assignment*: _quality metric threshold_].
 
 *Application Note {counter:remark_count}*:: ST author may specify quality standards if the TOE follows such standard, or provide a description of the metrics used.
 
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -432,7 +432,6 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 *FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _an internal quality metric_, [*assignment*: _quality metric standard_]], using a threshold of [*assignment*: _quality metric threshold_].
 
-*Application Note {counter:remark_count}*:: ST author may specify quality standards if the TOE follows such standard, or provide a description of the metrics used.
 
 ==== Protection of the TSF (FPT)
 ===== FPT_BDP_EXT.1 Biometric data processing [[FPT_BDP_EXT.1]]
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -432,9 +432,9 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_MBV_EXT.2 Quality of biometric samples for biometric verification [[FIA_MBV_EXT.2]]
 
-*FIA_MBV_EXT.2.1* The TSF shall only use samples of sufficient quality to verify the user.
+*FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _no independent quality metric, [*assignment*: quality metric standard]_].
 
-*Application Note {counter:remark_count}*:: ST author may refine “sufficient quality” to specify quality standards if the TOE follows such standard.
+*Application Note {counter:remark_count}*:: ST author may specify quality standards if the TOE follows such standard, or provide a description of the metrics used.
 
 ==== Protection of the TSF (FPT)
 ===== FPT_BDP_EXT.1 Biometric data processing [[FPT_BDP_EXT.1]]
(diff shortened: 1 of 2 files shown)


gfiumara:
I think this needs more if we're changing it. Specifying the quality standard used (independent or proprietary) is one thing, but the evaluator still needs to know the treshold of that metric to ensure that samples below that threshold are rejected. For example, if some quality metric produces a value 0 (worst) - 100 (best), the author should indicate their threshold is X. The evaluator then knows that if they present a sample of quality <X, it should be rejected.

Instead of:
> As such, sample data shall have [*selection*: _no independent quality metric, [*assignment*: quality metric standard]_].

Would it make sense for something like:
> Sufficiency of sample data shall be determined by measuring sample with  [*selection*: _an internal quality metric, [*assignment*: quality metric standard]_], using a threshold of [*assignment*: _quality metric threshold_].

gfiumara:
Because of the additions in the supporting document, I don't think we need the application note at all. Removing it would also remove a normative _may_ that we now know we shouldn't have in a note.

n-kai (commented):
**If no independent quality metric is claimed, the TSS must include a desription of how quality is measured**

However the vendor may not disclose such detail in the TSS... That's my concern.

woodbe:
things to add here: 

- application note about "internal quality metric" and what it means
- define in the SD (and anywhere else needed) a confidential TSS (like the FE KMD) for proprietary info needed for the lab/scheme but which should not be published generally (do this as a new PR)

woodbe:
I am not certain if we should get more detailed. I picked these as the SFRs that would be most likely to use the BMD at this point:

FIA_MBE_EXT.2,1
FIA_MBE_EXT.3,1
FIA_MBV_EXT.1.2
FIA_MBV_EXT.2,1
FIA_MBE_EXT.3,1

If we add the new SFR as mentioned in #324 then I would expect that to be added here as well. But I do not know that any of this is actually mandatory for the vendor to use. If someone is OK with all this being public, it could just be in the TSS, but if not, then this is useful.

n-kai (approved):

n-kai:
I believe that BMD is necessary, especially for PAD that biometric vendors don't want disclose any information how the TOE detects PAI but the evaluator needs such information to conduct objective evaluation.

Protection Profile/BiocPP.adoc
```diff
@@ -432,9 +430,9 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_MBV_EXT.2 Quality of biometric samples for biometric verification [[FIA_MBV_EXT.2]]
 
-*FIA_MBV_EXT.2.1* The TSF shall only use samples of sufficient quality to verify the user.
+*FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _no independent quality metric, [*assignment*: quality metric standard]_].
```
gfiumara, line 433: * independent -> internal
 * add threshold

```suggestion
*FIA_MBV_EXT.2.1* The TSF shall only use biometric samples of sufficient quality for verification. As such, sample data shall have [*selection*: _an internal quality metric_, [*assignment*: _quality metric standard_]], using a threshold of [*assignment*: _quality metric threshold_].
```

Supporting Documents/BS_SD.adoc
```diff
@@ -326,6 +326,8 @@ Following input is required from the developer.
 
 [loweralpha]
 . TSS shall explain how the TOE meets FIA_MBE_EXT.2 at high level description
+** If an independent quality metric is claimed, the TSS must include information about the metric
+** If no independent quality metric is claimed, the TSS must include a desription of how quality is measured
```
gfiumara, line 330: * I think these two bullets say the same thing
 * independent -> internal

```suggestion
** If a non-standard (i.e., internal) quality metric is claimed, the TSS must include information about the metric
```

Supporting Documents/BS_SD.adoc
```diff
@@ -422,6 +424,8 @@ Following input is required from the developer.
 
 [loweralpha]
 . TSS shall explain how the TOE meets FIA_MBV_EXT.1 at high level description
+** If an independent quality metric is claimed, the TSS must include information about the metric
+** If no independent quality metric is claimed, the TSS must include a desription of how quality is measured
```
gfiumara, line 428:
```suggestion
** If a non-standard (i.e., internal) quality metric is claimed, the TSS must include information about the metric
```

Protection Profile/BiocPP.adoc
```diff
@@ -432,9 +430,9 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
...
 
-*Application Note {counter:remark_count}*:: ST author may refine “sufficient quality” to specify quality standards if the TOE follows such standard.
+*Application Note {counter:remark_count}*:: ST author may specify quality standards if the TOE follows such standard, or provide a description of the metrics used.
```
gfiumara, line 435: This application note can be removed (was removed for enrolment quality).
```suggestion
```

Supporting Documents/BS_SD.adoc
woodbe: Is it sufficient to only point out expectations if an internal metric is used? Is it clear that an independent standard should also be noted without stating it here?

gfiumara: I may have misunderstood a reporting requirement:
 * If a documented standard quality metric is used, I think saying the standard name + threshold (lines 406, 433) is sufficient. The evaluator should be able to refer to the standard for more information.
 * If a proprietary quality metric is used, information about the proprietary metric should be required so the evaluator can attempt to make samples of different quality.

woodbe: Actually I think these should just be removed and different text added later in the description (looking at it). I don't think anything is needed here, or if anything is, it should just state that the quality metric description should be included, and leave it vague since this is actually like an overview section. It needs to be included later in some more detail (not much more, but under the assessment strategies section)

woodbe:
resolving the quality metrics expectations will need to be handled as part of #310

PR #320, opened 2021-04-16
updates for FIA_MBV_EXT.1.2

### LINKED ISSUE #306: [NIAP] FIA_MBV_EXT.1.2 application notes
Section 8.2.3, FIA_MBV_EXT.1.2, application note 14: This may be more of a endorsement statement thing, but it seems that the ST author is free to put anything they want into the assignments that essentially deal with how well the TOE works. This is not desirable from a "service-to-our-users" perspective. If two TOEs have a certificate and one TOE complies with NIST recommendations on values and the other does not, do we really expect the user to read in the ST to find these values and then make a determination of goodness? The goal for PPs is setting a clear minimum bar that the community believes is both achievable and provides some value to the end user.

Section 8.2.3, FIA_MBV_EXT.1.2, application note 14, item c: This item has the general problem mentioned earlier of containing "shall" statements even though 1) it's in an application note and 2) the introductory text in the application note says "...the ST author should consider..."

The more important issue is that this item uses the term "target error rates", but that is not defined in the glossary (nor anywhere else in the PP-Module), nor is it defined in the SD. It's unclear what this means, especially when the last sentence says that the SD describes how to objectively evaluate these things that aren't defined anywhere (and in fact, the SD does not do this).

woodbe:
This is to close #306.

This sets minimum values for the FAR/FRR and removes the "target"

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -411,7 +411,7 @@ This section lists SFRs for the biometric enrolment and verification.
 
 *FIA_MBV_EXT.1.1*:: The TSF shall provide a biometric verification mechanism using [*selection*: _fingerprint, eye, face, voice, vein_, [*assignment*: _other modality_]].
 
-*FIA_MBV_EXT.1.2*:: The TSF shall provide a biometric verification mechanism with the [*selection*: _FMR, FAR_] not exceeding [*assignment*: _defined value_] and [*selection*: _FNMR, FRR_] not exceeding [*assignment*: _defined value_].
+*FIA_MBV_EXT.1.2*:: The TSF shall provide a biometric verification mechanism with the [*selection*: _FMR, FAR_] not exceeding [*assignment*: _a value equal to or less than to 1:1000_] and [*selection*: _FNMR, FRR_] not exceeding [*assignment*: _a value equal to or less than 10%_].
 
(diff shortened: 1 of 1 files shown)


n-kai (commented):
According to the NIAP **The goal for PPs is setting a clear minimum bar that the community believes is both achievable and provides some value to the end user**

Is this "a value equal to or less than to 1:1000" acceptable **minimum bar** for NIAP for biometric verification performance?

woodbe:
The MDF currently lists 1:100 as the minimum, though I'm not exactly sure how that works given their SAFAR calculations.

n-kai:
See my comment in #306. It's OK to remove "target" if this term is not used by mobile biometric vendors but I need more time to define **minimum bar**.

woodbe:
I don't really know about the "target". I don't think it is needed as I read the text, having it there or not I don't see a change in the meaning. I don't think it needs to be defined, either, since it is clearly what you are trying to achieve. My concern with adding the definition is that we probably then need to review the full text to see if that should be added in more places vs just removing it in the small number where it is currently shown. I think that is the bigger thing, it's use seemed limited to just this requirement in a few sentences and not in any of the descriptive text at the beginning (or elsewhere).

n-kai:
@woodbe 
I agree to remove "target" (I just wanted to explain why I used the term "target")

woodbe:
Closing as superseded by#336

PR #327, opened 2021-05-19
BMD descriptions

### LINKED ISSUE #325: BMD descriptions
Add explicit splits between TSS and BMD for assessments

woodbe:
This is to close #325

Proposed changes for all the current SFRs that could use the BMD.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -429,7 +429,7 @@ Following input is required from the developer.
 [loweralpha]
 . TSS shall explain how the TOE meets FIA_MBV_EXT.1 at high level description
 . BMD shall provide information about how the verification rates are tested
-** The CMD may refer to the developer's <<performance report>>
+** The BMD may refer to the developer's <<performance report>>
 . AGD guidance shall provide clear instructions for a user to verify one's biometric to unlock the computer
 . Supplementary information (developer’s [#performance report]#performance report#) shall describe the developer’s performance test protocol and result of testing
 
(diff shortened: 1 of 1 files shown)


gfiumara:
Do we need to be specific about the "details" that should be in the BMD for either PAD or quality? I'm not sure what specifically we could say. Is it sufficient to say something like "this technique employs our trained machine learning model for assessing quality, based on a dataset of 3000 images whose quality labels were applied by certified examiners"?

woodbe:
@gfiumara so given that we are providing a specific document which would not be publicly released, the expectation is that they will need to provide specific details that could be used to evaluate whether the testing is acceptable. The main point for the BMD notice in the docs is to specify that some level of overview must be provided in the public TSS so that someone who reads the published ST would be able to at least have a high level overview of how the product meets the requirements (or how the testing was done to show it). The BMD description would still have to be enough to actually pass the validator review, and there I would prefer not to get too prescriptive (though we have tried to provide outlines about what is expected).

So really the point of what is said is to specifically state you MUST add something to the TSS, and the rest can go into the BMD, instead of letting the vendor put everything into the BMD.

n-kai (approved):
LIN 432 The CMD? may refer -> The BMD may refer

gfiumara (approved):

Supporting Documents/BS_SD.adoc
```diff
@@ -427,7 +428,8 @@ Following input is required from the developer.
 
 [loweralpha]
 . TSS shall explain how the TOE meets FIA_MBV_EXT.1 at high level description
-** If a non-standard (i.e., internal) quality metric is claimed, the TSS must include information about the metric
+. BMD shall provide information about how the verification rates are tested
+** The CMD may refer to the developer's <<performance report>>
```
gfiumara, line 432:
```suggestion
** The BMD may refer to the developer's <<performance report>>
```

PR #329, opened 2021-05-19
secure transmission by trusted path

### LINKED ISSUE #324: "secure" sensor input SFR
I was thinking about our SFRs and the changes we are making for the MDF, and one thing that I thought of was that we don't have some sort of secure path between the input sensor and the processing. This would probably be something along the lines of:

> The TOE shall prevent the operating system on the device from intercepting biometric input from the user.

This could be a new FPT_BDP_EXT.2 requirement.

The idea would be that the OS must not have access to the path between the sensor and the separate execution environment, and hence be able to possibly copy, modify or replace the data submitted by the user.

I had largely been relying on the concept of the SEE that we had defined to handle this, but now that we have been editing for the MDF and taken that away, I don't see how this trusted path is established/verified, and so I think we may need a new requirement for it.

woodbe:
This is to close #324 by adding FTP_TRP.1 based on the GP TEE BS PPM.

The ATE_IND testing is not completed.

After doing most of this I am not sure this is the right path to go down, but I am providing it anyway as a possible solution.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -99,7 +99,7 @@ Presentation Attack Detection (PAD)::
 (Biometric) Sample::
 	User’s biometric measures as output by the data capture subsystem of the TOE.
 Separate Execution Environment (SEE)::
-	An operating environment separate from the main Computer operating system. Access to this environment is highly restricted and may be made available through special processor modes, separate security processors or a combination to provide this separation.
+	An operating environment separate from the main computer operating system. Access to this environment is highly restricted and may be made available through special processor modes, separate security processors or a combination to provide this separation.
 Similarity Score::
 	Measure of the similarity between features derived from a sample and a stored template, or a measure of how well these features fit a user’s reference model.
 Template::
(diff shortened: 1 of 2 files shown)


n-kai (approved):
I am also digging into this issue and think that FTP_TRP is a good choice but I will probably add more to assessment strategy in BIOSD.

Protection Profile/BiocPP.adoc
```diff
@@ -452,14 +452,25 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 ==== Protection of the TSF (FPT)
 ===== FPT_BDP_EXT.1 Biometric data processing [[FPT_BDP_EXT.1]]
 
-*FPT_BDP_EXT.1.1*:: Processing of plaintext biometric data used to generate templates and perform sample comparisons shall be hardware-isolated from the OS on the TSF in runtime.
+*FPT_BDP_EXT.1.1*:: Processing of plaintext biometric data used to generate templates and perform sample comparisons shall be hardware-isolated from the main computer operating system on the TSF in runtime.
```
The-Fiona, line 455: Sorry, since I have been "absent" for a while and have missed any discussions..
Could we consider "platform operating system" rather than "main computer operating system"? It might be a clearer term, especially for MDF PP and GPOS PP.

Supporting Documents/BS_SD.adoc
```diff
@@ -657,6 +657,48 @@ The evaluator can pass this EA only if the evaluator confirms that:
 
 The evaluator shall report the summary of result of EA defined above, especially how the evaluator reaches the Pass/Fail judgement based on the Pass/Fail criteria.
 
+=== Trusted Path/Channels (FTP)
+==== EA for FTP_TRP.1
+===== Objective of the EA
+The path between the biometric capture sensor and the SEE where processing of the biometric characteristics will be performed per <<FPT_BDP_EXT.1>> or interactions to the main computer operating system occur (such as providing the results of a match), must protect the captured data from being tampered with or duplicated in transit. The evaluator shall confirm that this data path is protected from access by untrusted components (generally defined as anything in the main computer operating system).
```
The-Fiona, line 663: Should "match" be "comparison"? (See above change (Line 868)

Supporting Documents/BS_SD.adoc
```diff
@@ -657,6 +657,48 @@ The evaluator can pass this EA only if the evaluator confirms that:
...
+[loweralpha]
+. All TSF modules use a trusted path to transmit any data between components and the SEE and any entities outside the SEE including the main computer operating system can’t interfere with processing of these modules
+
+The evaluator shall keep in mind that the objective of this EA is not evaluating the SEE itself. This EA is derived from ASE_TSS.1.1 which requires that the TSS to provide potential consumers of the TOE with a high-level view of how the developer intends to satisfy each SFR. The evaluator shall check the TSS and BMD to seek for a logical explanation how the above criteria are satisfied considering this scope of the requirement.
+
+====== Strategy for ATE_IND
+TODO
+
+===== Pass/Fail criteria
+The evaluator can pass this EA only if the evaluator confirms that:
```
The-Fiona, line 693: "can" => "shall" ?

Protection Profile/BiocPP.adoc
woodbe: @The-Fiona so it could, but it happened to be that "main computer operating system" was how we defined the difference between say Android and the TEE (or iOS and the Security Enclave) when we had to redefine the SEE for the integration to MDF. We could make that change though, but it would have to be done in a lot of places, not just here.

Supporting Documents/BS_SD.adoc
gfiumara: Yes, per ISO/IEC 2382-37. The "result of a match" could be a "comparison decision."

woodbe:
This is being closed as the FDP_BDP_EXT expansion was used instead.

PR #330, opened 2021-05-19
Expansion of FPT_BDP_EXT

### LINKED ISSUE #324: "secure" sensor input SFR

woodbe:
This is to close #324

I think this is a better method than #329 that uses FTP_TRP in how we have defined the environment. It adds a new FPT_BDP_EXT.1.2 which specifically calls out internal transmission of biometric data between components (i.e. the sensor and the SEE)

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -99,7 +99,7 @@ Presentation Attack Detection (PAD)::
 (Biometric) Sample::
 	User’s biometric measures as output by the data capture subsystem of the TOE.
 Separate Execution Environment (SEE)::
-	An operating environment separate from the main Computer operating system. Access to this environment is highly restricted and may be made available through special processor modes, separate security processors or a combination to provide this separation.
+	An operating environment separate from the main computer operating system. Access to this environment is highly restricted and may be made available through special processor modes, separate security processors or a combination to provide this separation.
 Similarity Score::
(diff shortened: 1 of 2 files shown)


n-kai (approved):
I agree with this direction with #334

woodbe:
@n-kai I am worried about adding FPT_AEX_EXT.4.3 to the MDF. While it makes sense, I don't know that I would see this as a refinement of an existing requirement. The MDF basically dances around the requirement that there needs to be something separate from the main OS (as seen before, the only mentions I could come up with that point to it are that the REK needs to be isolated and some vague mumbo-jumbo about protecting keys). 

This could be something that we push for as an MDF 3.3 update itself, but we would need to talk with NIAP about that.

I'm good with the FCS_CKM app note.

I guess we could go this way and see what NIAP says, holding the FTP_TRP option as a backup.

gfiumara (approved):

PR #334, opened 2021-06-09
FPT_BDP_EXT_change

n-kai:
I agree with FPT_BDP_EXT with the following changes.

* Application note of FPT_BDP_EXT.1.1 is moved to one of FCS_CKM_EXT.4 because it's about the destruction of biometric data and I see very similar description for REK there.

* Add FPT_AEX_EXT.4.3 to modified SFR because, as NIAP commented, mobile device itself have to provide the SEE.

* Modify FPT_BDP_EXT based on introduction of FPT_AEX_EXT.4.3

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -385,6 +385,20 @@ TODO: Need to add additional EAs for modified and additional SFR to <<BIOSD>>
 
 The SFRs listed in this section are defined in the <<PP_MD_V3.3>> and relevant to the secure operation of the biometric enrolment and verification. It is necessary for the ST author to complete selections and/or assignments for these SFRs in a specific manner in order to ensure that the functionality provided by the mobile device is consistent with the functionality required by the biometric enrolment and verification in order for it to conform to this PP-Module.
 
+===== FCS_CKM_EXT.4 Key Destruction [[FCS_CKM_EXT.4]]
+
+*Application Note {counter:remark_count}*:: This SFR is functionally identical to what is defined in the <<PP_MD_V3.3>>. However, the following note should be added to the Application note of FCS_CKM_EXT.4. 
++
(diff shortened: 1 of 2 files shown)


woodbe:
@n-kai I am worried about adding FPT_AEX_EXT.4.3 to the MDF. While it makes sense, I don't know that I would see this as a refinement of an existing requirement. The MDF basically dances around the requirement that there needs to be something separate from the main OS (as seen before, the only mentions I could come up with that point to it are that the REK needs to be isolated and some vague mumbo-jumbo about protecting keys).

This could be something that we push for as an MDF 3.3 update itself, but we would need to talk with NIAP about that.

I'm good with the FCS_CKM app note.

I guess we could go this way and see what NIAP says, holding the FTP_TRP option as a backup.

n-kai:
@woodbe
The reason why I added FPT_AEX_EXT.4.3 is that adding this requirement can address the following NIAP comment in a better way than FTP_TRP option. But I agree that we would need to talk with NIAP about this. 

**Section 8.3, FPT_BDP_EXT.x: These requirements are written with respect to a Security Execution Environment. Since this is a PP-Module, there should be some verification that the Base PP requires a Security Execution Environment--in this way, the assurance activities can be written in a more objective way.**

woodbe:
@n-kai So that is why I re-wrote the secure execution environment into the separate execution environment as that matched with some of the language in the MDF. I think that FPT_AEX_EXT.4 could be used for this as it is with some expansions to the app notes. The biometric system is really composed of software and the sensor that should be separate, so maybe the one thing that isn't noted in the app note is that some hardware may need to have restricted access as well. Overall though, I can see getting what we need out of the requirement as it is written, as long as it is understood how to interpret the biometrics system into what is there.

I'm inclined to not have the new SFR for the public review, and add a note that this is a question to be specifically answered (and maybe create a specific issue for it to point the reviewers to for discussion).

n-kai:
@woodbe 
OK, I will remove FPT_AEX_EXT.4.3 and add some text to App note.

n-kai:
@woodbe 
Thank you. You have already done by #335.

PR #335, opened 2021-06-09
Changes to have less MDF changes

woodbe:
This tries to make the application note edits more explicit by copying in the paragraph from the MDF and adding edits to it here (I don't know if that is the right way or if NIAP would prefer something else).

I have removed the new FPT_AEX_EXT4.3 requirement by adjusting the app note to have more explicit notes about separate processor modes or even processors so it is clear these can be part of the description. We can suggest an SEE concept to NIAP for the revision where we are linked, but I think we should minimize drastic changes from our side for that. So I have tried to define it within. I think this should work, but maybe requires some more definition as to the expected description about what we would want to see (probably under the SD which I have added these into, but just by pointing them to the original EA).


n-kai (changes requested):
**the** in Line 393 should be deleted
The **the** Application Note following FCS_CKM_EXT.4.2

I don't have any other comment for this PR and will approve this after fixing the above error.

n-kai (approved):

PR #344, opened 2021-06-15
Edit to point to quality papers

### LINKED ISSUE #339: Point to quality paper vs ask vendor
https://github.com/biometricITC/cPP-biometrics/blame/b9a70e87cfe401d0beb66a60ab01884e999a2676/Supporting%20Documents/BS_SD.adoc#L520

I am thinking that maybe this should point to a quality paper instead of telling the evaluator to ask the vendor about creating low quality samples. The point of adding the paper was to provide guidance about quality, so it would seem that based on that the evaluator should be able to try. I don't have an issue stating that if you can't, then ask, but I would expect that based on the documentation about samples the evaluator should be able to do this.

### LINKED ISSUE #341: Vein quality reference
I realized while looking at the quality section that we don't have something specifically for vein. I found this link which I think will probably work, but am not certain:

https://link.springer.com/chapter/10.1007/978-3-030-27731-4_1#Sec21

I think we should add this to the list of references and then set mentions to the quality docs to point to both the current one as well as this one.

If there is a better reference though I would definitely take it.

woodbe:
This is to add a reference to review the quality papers for how to create low quality samples, and if that is unsuccessful, to ask the vendor (instead of asking first).

I have also added a reference related to vascular state of the art.

This will close #339 and close #341

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -517,7 +517,7 @@ The following test steps require the developer to provide access to a test platf
 +
 . The evaluator may examine feedback messages shown (e.g. “Finger too wet” in case of fingerprint verification) to understand how the evaluator can present a biometric characteristic that does not meet the <<MBE assessment criteria for samples, assessment criteria for samples>>
 
-. The evaluator shall review public papers on biometric quality (see <<Related Documents>> for some referenced quality papers) for the appropriate modality to determine how to create low quality samples if the TOE doesn't show any feedback messages
+. The evaluator shall review public papers on biometric quality (see <<Related Documents>> for some referenced quality papers) for the appropriate modality to determine how to create low quality samples. The evaluator shall document the papers used for reference in the evaluation report.
 .. The evaluator may ask the developer for assistance if public methods are unsuccessful in creating the low quality samples
 
 . The evaluator shall present biometric samples of acceptable quality for biometric verification that satisfy the assessment criteria described in <<MBV assessment criteria for samples, assessment criteria for samples>>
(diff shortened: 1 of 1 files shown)


Supporting Documents/BS_SD.adoc
```diff
@@ -517,7 +517,8 @@ The following test steps require the developer to provide access to a test platf
 +
 . The evaluator may examine feedback messages shown (e.g. “Finger too wet” in case of fingerprint verification) to understand how the evaluator can present a biometric characteristic that does not meet the <<MBE assessment criteria for samples, assessment criteria for samples>>
 
-. The evaluator may ask the developer how low quality samples can be created if the TOE doesn’t show any feedback messages
+. The evaluator shall review public papers on biometric quality (see <<Related Documents>> for some referenced quality papers) for the appropriate modality to determine how to create low quality samples. The evaluator shall document the papers used for reference in the evaluation report.
+.. The evaluator may ask the developer for assistance if public methods are unsuccessful in creating the low quality samples
```
gfiumara, line 521: Quality is relative to the algorithm used to measure it. While there's some general criteria for most modalities, they might not reflect a low quality sample for a particular quality algorithm. Below might help?

```suggestion
. The evaluator shall review public papers on biometric quality (see <<Related Documents>> for some referenced quality papers) for the appropriate modality and algorithm to determine how to create low quality samples. The evaluator shall document the papers used for reference in the evaluation report.
.. The evaluator may ask the developer for assistance if public methods are unsuccessful in creating the low quality samples (e.g., the quality method is not public).
```

gfiumara (approved):

gregott (approved):

ccolin318 (approved):

PR #345, opened 2021-06-15
New types of tests

### LINKED ISSUE #342: New artefacts test plan
https://github.com/biometricITC/cPP-biometrics/blame/b9a70e87cfe401d0beb66a60ab01884e999a2676/Supporting%20Documents/BS_SD.adoc#L904

So we originally wrote this with an expectation that a lab may add a new set of tests and use them, pretty much as part of the evaluation and then they would need to submit them. Given that we wrote a process for adding new procedures to an existing toolbox, I think we should remove this section from the BIOSD and replace it with a note that the lab should look at the beginning of the evaluation for new tests that aren't in the current toolbox, and if so, work to add them before starting the eval since they won't be allowed until approved.

We could take the information in this section and put it into the toolbox update process as a more fully described set of expectations as to what is expected when creating the new tests.

woodbe:
This is a start to the new artefacts test plan.

This will close #342

This update changes the assumptions section in the text. Other changes still need to be updated.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -913,10 +913,10 @@ For enhanced security that is easy to use, the TOE implements biometric verifica
 Attacker may use any tools or materials that are normally available at home and normal office environment such as laptop PC or office printer to attack the TOE. Attacker may also use any services (e.g. printing services to print a high-resolution photo of target users to create a face artefact) if such services are available at low cost.
 
 [loweralpha, start=2]
-. Assumptions in <<BIOPP-Module>>
+. Assumptions and Objectives in <<PP_MD_V3.3>>
 +
 --
-<<BIOPP-Module>> defines *A.User* and the evaluator shall assume that the computers are configured securely by users. The evaluator shall make the following assumptions:
+<<PP_MD_V3.3>> defines *A.CONFIG* and the evaluator shall assume that the computers are configured securely to enforce the TOE security policies and *A.PROPER_USER* where the evaluator shall assume that the user is not negligent or hostile. The evaluator shall make the following assumptions:
 
 [arabic]
 .. A user has enroled following guidance provided by the TOE
@@ -924,7 +924,7 @@ Attacker may use any tools or materials that are normally available at home and 
 +
 For efficiency, the evaluator can increase the maximum number of unsuccessful biometric authentication attempts to conduct the testing. However, as the computer shall be evaluated in the evaluated configuration, any attack needs to succeed within the allowed number of biometric authentication attempts defined in the ST to be considered a successful attack.
 
-<<BIOPP-Module>> also defines *A.Protection* and the evaluator shall assume that biometric data is adequately protected. Especially the evaluator shall make the following assumptions:
+<<PP_MD_V3.3>> also defines *O.INTEGRITY* and the evaluator shall assume that biometric data is adequately protected. Especially the evaluator shall make the following assumptions:
 
 [arabic, start=1]
 .. Attacker can not access the results of PAD subsystem, so they can not tune the artefacts based on the PAD score


woodbe:
I moved the new artefact text that we will want to use in the toolbox process into the biometricITC/Administration#41 pull request to update that document with more details on the expectation of the changes.

gfiumara (approved):

n-kai (changes requested):
See #350 for my proposed change to your PR.

iTC is responsible to maintain the BIOPPM, BIOSD and Toolbox but iTC can't approve or reject evaluators' action (i.e. finding new artefact and test it during the testing).  It's certifier's job.

Supporting Documents/BS_SD.adoc
```diff
@@ -891,57 +891,16 @@ This Section describes EAs for AVA_VAN.1 step by step following the order of AVA
 
 The evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
```
gfiumara, line 892: Noticed this while reading @n-kai's PR:

> New artefact species are those artefacts that are out of scope

This may need to be reworded, since "out of scope" also includes non-basic attack potential artefacts. Artefacts that are "significantly different" from or made using significantly different materials but still basic attack potential.

woodbe:
@n-kai While I agree that the iTC does not have the ability to control the evaluators actions, as we are the ones in control of the Toolbox, I don't see how it can be acceptable to allow someone to add new tests to the toolbox and then say that it is comparable to an evaluation that didn't have those.

My concern is something along these lines:

- vendor 1 tests to face v1
- the lab for vendor 2 decides there are new tests that should be done, and performs them, but does not add them to the toolbox, so still claiming support to the same face v1
- after the evaluation, the lab submits the new tests to the iTC, and we determine they should be changed, so the way they were run is not acceptable

So now vendor 2 has run the tests that meet v1, but also a set of tests that don't meet anything in v1 or anything in the approved v2 from the iTC. How would this be handled? How would this be repeatable (the NIAP requirement)?

This is why I want to say that the process for adding a new toolbox artefact would run something like this:

1. lab presents full set of docs for new artefacts to iTC (using toolbox TD process I will write up)
2. iTC will take 2 weeks to review the docs, and probably make some suggestions (likely working with the lab on it)
3. docs are added to the toolbox and version is iterated/published
4. full normal review proceeds on artefact addition, which could lead to no changes or additional updates to the docs. if changes are made, another revision is published, on whatever schedule the iTC takes

So this would mean going from v1, the lab would provide updates that we would use to create a v2, which would be accepted as the iteration for that modality. If the iTC decides additional changes are needed, then a v3 would be created, but v2 would be able to be used for the evaluation, and would be in effect as the current set of tests for all new evals running (so if another vendor started 3 months later, they would also have to use these docs with the new artefacts.

What I am trying to avoid is having 2 vendors testing the same modality and one of these scenarios happening (because the lab can do what it wants and still be approved):

- one lab decides that there are new artefacts to be used while the second one does not (in this case the second one happening a month or 2 after the one that did propose the changes)
- 2 labs each see the new artefact type and create individual plans that are different, and so while both vendors test against the new artefacts, they don't do so in the same manner, so the evaluations aren't technically equivalent

The way less ideal way to handle this scenario, I would think, would be to to say that until the new artefacts are approved, even if there is testing to the new artefacts, the certificate can only claim compliance to the current published version of the toolbox. If the scheme wants to list additional details about tests that is fine, but there wouldn't be any compliance claim to anything except the current version of the toolbox. The problem here though is that there is no consistency for any new artefacts, and likely this will be frowned upon (especially by NIAP).

Supporting Documents/BS_SD.adoc
woodbe: I have updated the text to remove the out of scope phrase. I think this should work well.

n-kai (approved):

woodbe:
I have added another line based on this [comment](https://github.com/biometricITC/cPP-biometrics/pull/350#issuecomment-869661720) while merging that PR.

PR #347, opened 2021-06-16
IAPAR update

### LINKED ISSUE #343: IAPAR defined in SFR
As we have been adding/adjusting the FAR/FRR claims to be more explicit, I am thinking we should probably do the same for the MBE/MBV_EXT.3 requirements. 

FIDO uses the IAPAR (Impostor Attack Presentation Accept Rate) from ISO-30107-3 (we have part 1 listed in the PPM and part 3 in the SD) and they specify a limit of 7% (though in the upcoming version they will actually relax that some to 15% as the minimum to pass with an option to get to 7% and have it listed on the certificate).

Given that we have set a lower acceptable boundary for the FAR/FRR values, it would seem that we need to do so here as well. This would entail rewriting the SFRs though I do not think the SD will need much change.

So one question is what should the minimum be? I would vote for the 15% as the highest failure rate that is allowed. I don't know if we also need to specify confidence intervals here as well (given what this testing is, I'm not sure that is applicable).

woodbe:
This is to close #343

One concern I have here is that I updated FIA_MBE_EXT.3.1 to match, but technically IAPAR is for verification, so I'm not sure we can use that for enrolment. I don't know what to do if we don't, since then the enrolment requirement is vague, but I also don't know if an IAPAR for verification is in any way equivalent for enrolment.

I didn't make any further changes in terms of tables or such that would explain the calculations or anything differently (or more explicitly), so we may end up needing to do something along those lines, but I'm not sure.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -88,8 +88,6 @@ False Reject Rate (FRR)::
 	Digital representation of the information extracted from a sample (by the signal processing subsystem) that will be used to construct or compare against enrolment templates.
 Imposter Attack Presentation Accept Rate (IAPAR)::
     In a full-system evaluation of a verification system, proportion of impostor presentation attacks using the same artefact type that result in a accept.
-Imposter Attack Presentation Accept Rate for enrolment (IAPARE)::
-    In a full-system evaluation of a enrolment system, proportion of impostor presentation attacks using the same artefact type that result in a successful enrolment.
 Locked State::
 	Powered on Computer, with most functionalities unavailable for use. User authentication is required to access full functionality.
 (Biometric) Modality::
@@ -662,7 +660,7 @@ ST authors are free to choose none, some or all SFRs defined in this Section. Ju
 
 ==== FIA_MBE_EXT.3 Presentation attack detection for biometric enrolment [[FIA_MBE_EXT.3]]
 
-*FIA_MBE_EXT.3.1*:: The TSF shall provide a biometric enrolment mechanism with the IAPARE not exceeding [*assignment*: _value equal to or less than 15% (15:100)_] to prevent use of artificial presentation attack instruments from being successfully enroled.
+*FIA_MBE_EXT.3.1*:: The TSF shall prevent use of artificial presentation attack instruments from being successfully enroled.
 
 ==== FIA_MBV_EXT.3 Presentation attack detection for biometric verification [[FIA_MBV_EXT.3]]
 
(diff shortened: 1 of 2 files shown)


gfiumara:
For enrolment, you maybe want "bona fide presentation classification error rate (BPCER): _proportion of bona fide presentations incorrectly classified as presentation attacks in a specific scenario_," and/or the inverse, "attack presentation classification error rate (APCER): _proportion of attack presentations using the same PAI species incorrectly classified as bona fide presentations in a specific scenario_."

n-kai (changes requested):
I checked the latest ISO 30107-3 that is under revision. you can't use IAPAR for enrollment so new term something like below should be defined and this new term should be used in FIA_MBV_EXT.3.1

**Imposter Attack Presentation Accept Rate for enrollment (IAPARE)
In a full-system evaluation of a enrollment system, proportion of impostor presentation attacks using the same artefact type that result in a successful enrollement**

gfiumara:
I don't think I agree with creating a new term IAPARE that is not defined by ISO. Enrolment doesn't deal with a verification system at all, it would only deal with the "data capture subsystem," which is the term specified in the published ISO/IEC 30107-3:2017. The second A in IAPAR is about acceptance, formally M for match, but ISO/IEC depreciated using _match_ this way. This wouldn't follow for enrolment.

If anything, the term should be IAPE(nrollment)Rate, but I think that might as well be the _attack presentation acquisition rate (APAR)_ ("proportion of attack presentations using the same PAI species from which the data capture subsystem acquires a biometric sample of sufficient quality").

woodbe:
Another option would be to not specify anything for IAPAR for enrolment and go back to what it was before.

I will admit I read the comment wrong and thought this was in the draft ISO (when I was reading the comment), so having it there is to capture it. I am not certain how to handle the enrolment case, as it is definitely the most tricky since you don't have anything to compare with readily at that point.

n-kai:
I agree with "Another option would be to not specify anything for IAPAR for enrolment and go back to what it was before.". PAD for enrolment is optional requirement and ISO/IEC 30107-3 doesn't require to report the error rate like IAPAR.

woodbe:
I have removed the FIA_MBE_EXT.3 requirement changes to leave it as it was originally at this point. So the changes are now just the IAPAR on the FIA_MBV_EXT.3 requirement

n-kai (approved):

gfiumara (approved):

woodbe:
call this out for comment in the public review

woodbe:
agreed to be merged on the call

PR #348, opened 2021-06-17
Use APCER/BPCER

gfiumara:
Updates the IAPAR PR to use APCER/BPCER at some default value (no comment on if I think this value is too high) for enrolment. This couldn't be a suggestion on the PR because of adding the definitions at a location not available in the original PR.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -74,6 +74,10 @@ Computer::
 	A self-contained device which is composed of a hardware platform and its system software (operating system and applications). The device is typically some sort of general purpose computing platform, such as a laptop, tablet or smartphone that is designed to be portable (though this is not required). _In this version, the term Computer is used as a synonym for Mobile device. However, in the future version, this PP-Module will be updated to allow to use with the latest version of <<PP_OS>> and this italic text will also be removed._
 Computer User (User)::
 	The individual authorized to physically control and operate the Computer, usually the device owner. This person is responsible for configuring the TOE.
+Attack Presentation Classification Error Rate (APCER)::
+	Proportion of attack presentations using the same presentation attack instrument species incorrectly classified as bona fide presentations in a specific scenario.
+Bona Fide Presentation Classification Error Rate (BPCER)::
+	Proportion of bona fide presentations incorrectly classified as presentation attacks in a specific scenario.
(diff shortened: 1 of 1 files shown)


gfiumara:
On second thought, we're not actually asking these devices to tell us if a presentation is an attack or bona fide presentation, therefore the "classification rate" of presentations might be out of scope.

Another term from 30107-3 is "attack presentation acquisition rate (APAR): proportion of attack presentations using the same PAI species from which the data capture subsystem acquires a biometric sample of sufficient quality." Maybe that is more appropriate?

woodbe:
@gfiumara I'm assuming this is specifically then for enrolment, and not verification? This isn't a term I'm familiar with, so I'm fine going with it here, I had put in the IAPAR for the enrolment as I didn't know what else to do.

woodbe (approved):

woodbe:
@n-kai since I'm not as familiar with this, can you review this before I merge it into the IAPAR change I had made?

n-kai:
I checked the latest version of ISO/IEC 30107-3 that is under revision. APCER and BPCER are metrics for PAD subsystem. IAPAR is the one for full system. Our TOE corresponds to the full system that has PAD, quality check and matching functions. So we should use IAPAR instead of APCER and BPCER. See below for more information from 30107-3


attack presentation classification error rate
APCER 
proportion of attack presentations using the same PAI species incorrectly classified as bona fide presentations by a **PAD subsystem** in a specific scenario

bona fide presentation classification error rate
BPCER
proportion of bona fide presentations incorrectly classified as presentation attacks in a specific scenario

impostor attack presentation accept rate
IAPAR
in a **full-system** evaluation of a verification system, proportion of impostor attack presentations using the same PAI species that result in accept

**A PAD subsystem** is hardware and/or software that implements a PAD mechanism and makes an explicit declaration regarding the detection of presentation attacks. Results of the PAD mechanism are accessible to the evaluator and are an aspect of the evaluation.

**A full system** adds biometric comparison to the PAD subsystem or data capture subsystem, comprising a full end-to-end system. This leads to additional failure points for the PAI beyond PAD mechanisms and quality checks. In a full system, there might be one or multiple PAD mechanisms at different points in the system

woodbe:
So if we are looking at a full system then, it would seem we should close this pull request and focus on the IAPAR update.

PR #350, opened 2021-06-25
New artefact test modification

n-kai:
This PR will update the following text and make a few minor editorial update

_To utilize new artefacts, they must have approval from the BIO-iTC. The requirements for submitting new artefacts for use can be found at the Biometrics Security home page. If the new artefact type is approved, it will be included as part of the [Toolbox] for use. If the new artefact type is not approved, then only the artefacts already in the [Toolbox] may be used._

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -890,7 +890,7 @@ This Section describes EAs for AVA_VAN.1 step by step following the order of AVA
 ===== Search for new artefacts
 Each of the supported biometric modalities have a specific set of defined artefacts species in the <<Toolbox>> to be used in testing. These are devised based on publicly available information published by the publication date of the <<Toolbox>>. The BIO-iTC also verifies that test items cover all existing artefact species that are within the scope of Basic attack potential defined in <<Attack Potential and TOE resistance>>.
 
-However, new artefacts species may be found after the <<Toolbox>> is published. So the evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
+However, new artefacts species may be found after the <<Toolbox>> is published. The evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
 
 Those new artefact species that can be made by slightly modifying test items in the <<Toolbox>> are covered by the normal test plans.
 
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -894,7 +894,9 @@ The evaluator shall search publicly available information that is published afte
 Those new artefact species that can be made by slightly modifying test items in the <<Toolbox>> are covered by the normal test plans.
 
 ====== New artefact Toolbox updates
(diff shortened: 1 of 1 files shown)


Supporting Documents/BS_SD.adoc
```diff
@@ -888,16 +888,16 @@ The <<Toolbox>> describes a test subset and test documentation that is sufficien
...
 
-The evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
+However, new artefacts species may be found after the <<Toolbox>> is published. So the evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
```
gfiumara, line 893:
```suggestion
However, new artefacts species may be found after the <<Toolbox>> is published. The evaluator shall search publicly available information that is published after the publication date of the <<Toolbox>> to look for new artefact species. New artefact species are those artefacts that are out of scope of the <<Toolbox>> and need to be made in a completely different way with significantly different materials that are not covered by the <<Toolbox>>.
```

Supporting Documents/BS_SD.adoc
```diff
@@ -888,16 +888,16 @@ The <<Toolbox>> describes a test subset and test documentation that is sufficien
...
 
 ====== New artefact Toolbox updates
-To utilize new artefacts, they must have approval from the BIO-iTC. The requirements for submitting new artefacts for use can be found at the https://biometricitc.github.io/[Biometrics Security home page]. If the new artefact type is approved, it will be included as part of the <<Toolbox>> for use. If the new artefact type is not approved, then only the artefacts already in the <<Toolbox>> may be used.
-
-===== Identify candidate artefacts for testing
-Each of the supported biometric modalities have a specific set of defined artefacts to be used in testing. These are devised based on publicly available information published by the publication date of the <<Toolbox>>. The BIO-iTC also verifies that test items cover all existing artefact species that are within the scope of Basic attack potential defined in <<Attack Potential and TOE resistance>>. 
+The evaluator is strongly encouraged to report to the BIO-iTC when the evaluator can find the new artefact species that can successfully attack the TOE to add them to the <<Toolbox>>.
```
gfiumara, line 898: Can we require this instead?

I think we should also have the evaluator report new types of artefacts including those that cannot successfully attack the TOE because if different enough, they may be able to successfully attack other targets.

```suggestion
The evaluator shall report to the BIO-iTC when new artefact species are found so the artefacts may be added to the <<Toolbox>>. 
```

woodbe (approved):
My one concern on the new artefacts is that we are still implying that artefacts could be added before approval by the iTC. I think that it needs to be clear that while new artefacts can be added, and even provisionally added quickly (assuming they provide all the information), that if they don't have it approved by the iTC prior to testing, they cannot be considered part of the toolbox. I want this to be explicitly stated so it is not open to interpretation.

gfiumara:
@woodbe Can we address your concern by re-adding the deleted sentence?

 > To utilize new artefacts, they must have approval from the BIO-iTC.

or modifying it somehow, that would also incorporate some of the additional information you make [in this comment](https://github.com/biometricITC/cPP-biometrics/pull/345#issuecomment-868516037)?

 > To utilize new artefacts, they must have approval from the BIO-iTC **through  publication in an incremented toolbox version for the evaluator to evaluate against. [footnote: Publication can occur rapidly, typically ≤2 weeks]**

The conversation is split between the two PRs, but I think the concern is we want evaluators to be able to make new artefact if they want to, but not to claim them as official unless the iTC says so, especially in the event that the device does not pass the evaluation with the new artefact.

PR #349, opened 2021-06-17
Strict Guidance update

woodbe:
This initial commit isn't complete but is something to get started with. I plan to complete a second commit tomorrow that will finish off the changes (including all the references that now point to the assessment criteria for samples).

Comments on the parts that are done are fine, as are even additional commits.

This is to close #310

PR #351, opened 2021-06-27
Strict guidance mod

n-kai:
This PR was created to address an issue inputted by NK in #349 on 27th June.


gfiumara:
I'm under the impression that "internal quality metric" and "developer defined quality metric" are considered the same thing. I don't mind changing the terminology, but I want to make sure I'm not misunderstanding the significance of this change.

woodbe:
@gfiumara I think they are the same. Initially we had it noted as "independent" quality metric, and then agreed to change it to "internal". I don't think it matters what the term is as long as we agree and use it consistently.

woodbe (approved):
I am not in full agreement on the SD update, but will merge to consolidate the conversation and changed.

PR #353, opened 2021-09-06
Update for editorial fix in BIOSD

n-kai:
This PR address the comment No 1, 7 and 8 in my comments.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -71,7 +71,7 @@ Biometric Security international Technical Community (BIO-iTC)
 
 |1.1
 |TBD, 2021
-|Incorporated TDs and NIAP comments for PP_MD_V3.3 integration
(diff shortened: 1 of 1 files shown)


woodbe (approved):

woodbe:
A comment on the changes from PP_MD_V3.3 to PP_MD (also in #359).

While I get that the point is to remove some of the version references since the PP_MD may be updated and we would want to remain valid, the reality of that is that the PPM will need a TD-style update to show the current version (and this is just assuming that there is nothing else needed, like actual SFR changes or something else). So while this technically would make the number of changes smaller (since I wouldn't need to search and replace for PP_MD_V3.3 and replace it with PP_MD_V3.4), the references to the full PP will still need to be updated. I don't have an issue with this, I just don't think it provides a huge benefit beyond smaller-scoped TDs.

gfiumara (approved):

PR #355, opened 2021-09-06
Update performance testing in BIOSD

n-kai:
This PR addresses my comment No 3

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -412,11 +412,9 @@ The evaluator shall report the summary of result of EA defined above, especially
 
 ===== Objective of the EA
 
-The evaluator shall verify that the TOE implements the biometric verification mechanism whose error rates is equal or lower than the claimed error rates (i.e. value of FAR/FMR and FRR/FNMR specified in FIA_MBV_EXT.1.2).
+The evaluator shall verify that the TOE implements the biometric verification mechanism whose upper bound confidence interval of error rates don't exceed the claimed error rates (i.e. value of FAR/FMR and FRR/FNMR specified in FIA_MBV_EXT.1.2).
 
-The evaluator shall solely rely on the supplementary information (developer’s <<Developer’s performance report and its assessment strategy, performance report>>) to achieve this objective following instruction defined in Assessment Strategy.
-
-<<BIOPP-Module>> assumes that the biometric verification is not used for the security sensitive services and the TOE operational environment also limits the maximum number of failed verification attempts in succession. Therefore, risk of zero-effort impostor attempts is low and the developer may not follow the statistical method (e.g. Rule of 3 or Rule of 30) to measure the biometric verification performance.
+The evaluator shall solely rely on the supplementary information (developer’s <<Developer’s performance report and its assessment strategy, performance report>>) to achieve this objective following instruction defined in Assessment Strategy. <<BIOPP-Module>> assumes that the biometric verification is not used for the security sensitive services and the TOE operational environment also limits the maximum number of failed verification attempts in succession. Therefore, the evaluator does not need to take lots of time and cost to gather test subjects and conduct the independent testing for this SFR.
 
 ===== Dependency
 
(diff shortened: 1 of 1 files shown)


woodbe (approved):

Supporting Documents/BS_SD.adoc
```diff
@@ -411,11 +411,9 @@ The evaluator shall report the summary of result of EA defined above, especially
 
 ===== Objective of the EA
 
-The evaluator shall verify that the TOE implements the biometric verification mechanism whose error rates is equal or lower than the claimed error rates (i.e. value of FAR/FMR and FRR/FNMR specified in FIA_MBV_EXT.1.2).
+The evaluator shall verify that the TOE implements the biometric verification mechanism whose upper bound confidence interval of error rates don't exceed the claimed error rates (i.e. value of FAR/FMR and FRR/FNMR specified in FIA_MBV_EXT.1.2).
```
gfiumara, line 414:
```suggestion
The evaluator shall verify that the TOE implements the biometric verification mechanism whose upper bound confidence interval of error rates does not exceed the claimed error rates (i.e. value of FAR/FMR and FRR/FNMR specified in FIA_MBV_EXT.1.2).
```

gfiumara: We're using different confidence intervals depending on the error rate. Do we need to specify those everywhere, or can we refer to the section where they are defined (both here and elsewhere)? They aren't mentioned until an Example much later.

Supporting Documents/BS_SD.adoc
```diff
@@ -411,11 +411,9 @@ The evaluator shall report the summary of result of EA defined above, especially
...
-
-<<BIOPP-Module>> assumes that the biometric verification is not used for the security sensitive services and the TOE operational environment also limits the maximum number of failed verification attempts in succession. Therefore, risk of zero-effort impostor attempts is low and the developer may not follow the statistical method (e.g. Rule of 3 or Rule of 30) to measure the biometric verification performance.
+The evaluator shall solely rely on the supplementary information (developer’s performance report explained in <<Developer’s performance report and its assessment strategy>>) to achieve this objective following instruction defined in Assessment Strategy. The <<BIOPP-Module>> assumes that the biometric verification is not used for the security sensitive services and the TOE operational environment also limits the maximum number of failed verification attempts in succession. Therefore, the evaluator does not need to take lots of time and cost to gather test subjects and conduct the independent testing for this SFR.
```
gfiumara, line 416: I thought the older explanation was mostly fine and that it could be important to keep the "zero-effort impostor" term included in the explanation. However, if we keep this, let's mention that we mean a large quantity of people.

```suggestion
The evaluator shall solely rely on the supplementary information (developer’s performance report explained in <<Developer’s performance report and its assessment strategy>>) to achieve this objective following instruction defined in Assessment Strategy. The <<BIOPP-Module>> assumes that the biometric verification is not used for security sensitive services and the TOE operational environment also limits the maximum number of failed verification attempts in succession. Therefore, the evaluator does not need to gather large quantities of test subjects to conduct the independent testing for this SFR.
```

gfiumara:
Want to bubble up this comment for discussion because it was accidentally resolved in a typo fix by GitHub: We're using different confidence intervals depending on the error rate. Do we need to specify those everywhere we mention "upper bound confidence interval," or can we refer to the section where they are defined (both here and elsewhere)? They aren't mentioned until an Example much later.

n-kai:
@gfiumara 
Sorry but I can't understand your comment. FAR/FMR and FRR/FNMR are different variables and confidence interval should be estimated separately for each variable.  Do you mean that we should update the FIA_MBV_EXT.1.2 like as follows?

The TSF shall provide a biometric verification mechanism with the [selection: FMR, FAR] not exceeding [assignment: value equal to or less than 0.01% (1:104)] for the upper bound of [assignment: value equal to or greater than 80%] confidence interval **of [selection: FMR, FAR]** and, [selection: FNMR, FRR] not exceeding [assignment: value equal to or less than 5% (5:100)] for the upper bound of [assignment: value equal to or greater than 80%] confidence interval **of [selection: FNMR, FRR]**.

gfiumara:
> Sorry but I can't understand your comment. FAR/FMR and FRR/FNMR are different variables and confidence interval should be estimated separately for each variable. Do you mean that we should update the FIA_MBV_EXT.1.2 like as follows?

I was asking if the confidence interval (80%, 95%, etc) needs to be specified on line 449, for example. However, I realize now I pulled up the wrong branch when I glanced over that the PP and we do specify the CI there, so my comment is unnecessary (the selected CI _is_ already documented).

(Apologies for missing the meeting...some end of fiscal year meetings may conflict the next couple weeks)

gfiumara (approved):

PR #357, opened 2021-09-06
Update for PAD evaluation in BIOSD

n-kai:
This PR addresses my comment No 9


woodbe (approved):

gfiumara:
I want to be sure I'm not missing something, as I can't seem to find the original comment this PR is fixing: the pass/fail criteria is being removed here because it has been moved into the toolbox overview, correct?

woodbe:
@gfiumara that is my understanding. When I was looking at the earlier doc I couldn't quite see why this was located where it was (it seemed a little "off" in terms of placement), so it seems like we took that and put it into the toolboxes.

PR #361, opened 2021-09-30
Nkai update appnote for fia mbv ext.3.1

n-kai:
Add new app note that clarify IAPAR can be estimated based on attempt or transactions (FIDO defines the IAPAR based on number of transactions however BIOPPM allows to select attempt based error rate (FMR/FNMR) so attempt based IAPAR should also be allowed).

This PR also fix two types (PP_MD_V33) in BIOSD.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -655,7 +655,9 @@ ST authors are free to choose none, some or all SFRs defined in this Section. Ju
 
 *FIA_MBV_EXT.3.1*:: The TSF shall provide a biometric verification mechanism with the IAPAR not exceeding [*assignment*: _value equal to or less than 15% (15:100)_] to prevent use of artificial presentation attack instruments from being successfully verified.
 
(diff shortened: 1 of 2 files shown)


woodbe (approved):

gfiumara (approved):

ccolin318:
Is there a case where FMR/FNMR and FAR/FRR are selected at the same time in FIA_MBV_EXT.1.2? It seems that the PP allows this possibility. Do we need to worry about this?

woodbe:
@ccolin318 in my opinion, while that is technically possible, I don't think that will actually happen just given how the measurements are usually done. Usually the presentation of one or the other is mutually exclusive (pick one and that is what you use).

ccolin318:
@woodbe That makes sense to me. If a vendor does want to select both in the future, that is when we can make a change if necessary. But I don't think it is necessary at this time - I am adding my approval to this pull request.

ccolin318 (approved):

PR #366, opened 2021-11-12
Update of Interpretation branch

woodbe:


gfiumara (approved):

gfiumara:
Is there a way to avoid requiring approvals on bringing branches up to date? Not sure if destroying the branch and then re-creating it under the same name might be easier? So long as commits weren't squashed, the history should all be there.

woodbe:
@gfiumara Technically it should be a rebase, but I'm not as sure how to make that work (my more minimal GitHub knowledge here is part of the issue).

Issue #370, opened 2022-05-11
Supporting Document v1.1 FPT_BDP_EXT.1 Section 2.4.1.7.1 - Impossible Test!


xahun:
Supporting Document v1.1 FPT_BDP_EXT.1 Section 2.4.1.7.1 "a. Memory scan test" requires 1) the ability to essentially pause the TOE and dump a copy of memory, 2) resume normal operations after the dump, e.g., to perform a biometric function, and 3) the ability to pause and dump memory again.  I don't know about iPhones, but Android phones engineered to dump memory reboot after the dump, so this is not possible with any Android phones tested to date to my knowledge.  Note also that dumping might involve writing to an sdcard that lies under the phone battery, making this even more difficult for some TOE models.  Even if this capability were developed, the evaluator is faced with comparing upwards of 16GB of memory images with likely a very large number of changes given the time it necessarily takes to dump up to 16GB of memory externally.  Finally, how in the world would an evaluator determine if any memory changes "show access to biometric data"?  This test is essentially impossible given any reasonable resources.  The suggestion "If it is impractical or inadequate to conduct the following tests" of course then would always be true, but leaving it to the developer and evaluator to whip up and defend a test is not aligned with current CC direction - rather a reasonable test should be developed for this module.  I suggest working along the lines of FCS_CKM_EXT.4 where applicable internal/secret biometric data is logged before or while a biometric activity occurs, then the TOE memory is dumped and searched for those values.

ccparran:
I copied the SFRs:

FPT_BDP_EXT.1.1
Processing of plaintext biometric data shall be inside the SEE in runtime.
FPT_BDP_EXT.1.2
Transmission of plaintext biometric data between the capture sensor and the SEE shall be isolated from the main computer operating system on the TSF in runtime.

I am concerned that the changes proposed on the last call do not align with the SFR intentions. 

Section 2.4.1.1 in the BS_SD states: "the evaluator shall examine that the TOE processes any plaintext biometric data within the boundary of the SEE, and that the transmission of this data is via a channel protected from the main computer operating system. The SEE is responsible for preventing any entities outside the environment from accessing plaintext biometric data."

In the proposed step #2 "The developer tools shall write the biometric template data to an accessible location (either prior or during, depending on the type of test)", are we asking the evaluator to write to an accessible place other than the SEE, in order to conduct the test? Maybe the intent should be that the template data can not be found in memory because it is only written to the SEE?

[...]

PR #380, opened 2022-05-16
memory dump update

woodbe:
To respond to #370

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -605,10 +605,11 @@ The test is repeated for biometric enrolment and biometric verification (called 
 +
 [arabic]
 .. Using tools provided by the developer, the evaluator shall prepare the computer for a scan of the main computer operating system memory
-.. The evaluator shall take a memory scan before biometric transaction
-.. The evaluator shall perform the biometric transaction
-.. The evaluator shall perform a memory scan during the biometric transaction and compare the results to the scan before the biometric transaction
-.. The evaluator shall confirm that the changes between the two scans do not show access to the biometric data
+.. The developer tools shall write the biometric template data to an accessible location (either prior or during, depending on the type of test)
+.. The evaluator shall perform a biometric transaction
+.. The evaluator shall perform a memory scan during the biometric transaction
+.. The evaluator shall take a memory scan after the completion of a biometric transaction
+.. The evaluator shall confirm that the template data can only be found in memory during the biometric transaction
 [loweralpha]
 . TSFI invocation test
 +


n-kai (changes requested):
EA for the FCS_CKM_EXT.4 in the [PP_MDF] requests evaluators to do the similar memory scan for testing key destruction (See below). So our SD can also recommend evaluators to refer this EA to do our memory scan test so that evaluators can test both in the same way.

**EA for FCS_CKM_EXT.4**
_Record the value of the key in the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) subject to clearing.
    Cause the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) to perform a normal cryptographic processing with the key from Step #1.
    Cause the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) to clear the key.
    Cause the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) to stop the execution but not exit.
    Cause the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) to dump the entire memory of the [TOE](https://commoncriteria.github.io/pp/mobile-device/mobile-device-release.html#abbr_TOE) into a binary file.
    Search the content of the binary file created in Step #5 for instances of the known key value from Step #1.
    Break the key value from Step #1 into 3 similar sized pieces and perform a search using each piece._

So we can add the following text under a. 5. like follows.

_5. The evaluator shall confirm that the changes between the two scans do not show access to the biometric data_

_**The evaluator can refer to the EA for FCS_CKM_EXT.4 in the [PP_MDF] to do this memory scan test.**_

woodbe:
@n-kai This specific testing is what @xahun was pointing out is likely impractical for the biometric testing. The problem is how do you interrupt the testing in the middle of the sequence (I'm honestly not even sure how you would do that). The other concern I actually have is whether just checking for the template itself in memory or if it should be checking for the sensor input from the user (which wouldn't be known until after the user enters it, and then what are we looking for, the raw image, or the processed template used to compare, or something else)? 

In the end I think the problem here is that biometric systems do not line up neatly with entering a password. In that case I can simply dump the password and know that if that is found anywhere at some point it (that is outside of the bounds of use), then it is obviously a problem. But for a biometric, the template isn't going to look like the user input from a single scan, so what is it that I am checking for?

I can maybe see just saying to scan for something. For example, dump the template out and then try to scan for it after the sequence is complete (since the real point of the SFR is to show that it isn't available when no longer needed). It isn't clear that it can be done "in the middle" of the attempt, certainly by the user, and the cost of creating that debug capability vs what is shown after is not something I'm convinced is actually worthwhile.

woodbe:
@woodbe will proposed change to focus more on design of the biometric vs this test, and leave the testing to the TSFI if there are possible interfaces that could be used to export data outside the SEE.

woodbe:
The latest update is to remove the memory test as out of scope as it is testing the SEE boundary, and not the TOE boundary.

In line 597 there is a statement (which I left) about source code review being acceptable to ensure this as an alternative to testing (which I left specifically for the TSFI test). I don't want to make this a requirement for the general boundary issue (that everything is inside the SEE), but it is something to consider.

n-kai (approved):

PR #384, opened 2022-06-10
PAD optional

### LINKED ISSUE #375: Supporting Document v1.1 Section 7.2 ATE_IND.1
Supporting Document v1.1 Section 7.2 ATE_IND.1 - the identified tests are specifically for FIA_MBE_EXT.3 and FIA_MBV_EXT.3, so they should be moved.  As written it appears that those tests must always be performed, but based on statements found for those requirements, reference to those requirements in this section, and statements in the toolbox introduction it becomes clear these test activities go with these optional SFRs.  In any case, it should be very clear that this is optional and the conditions that would require it.

woodbe:
This is to close #375.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -847,7 +847,7 @@ The evaluator shall follow the same EAs defined in <<PP_MDF>> and the BIOSD. As 
 
 === Introduction
 
-PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if one (or both) of these requirements are included.
+PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if any of these requirements are included.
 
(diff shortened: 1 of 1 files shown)


Supporting Documents/BS_SD.adoc
```diff
@@ -847,6 +847,8 @@ The evaluator shall follow the same EAs defined in <<PP_MDF>> and the BIOSD. As
 
 === Introduction
 
+PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if one (or both) of these requirements are included.
```
gfiumara, line 850:
```suggestion
PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if any of these requirements are included.
```

gfiumara (approved):

Supporting Documents/BS_SD.adoc
gregott: That works for me.

gregott (approved):

n-kai (approved):

PR #385, opened 2022-06-10
Testing week justfication

### LINKED ISSUE #377: Supporting Document v1.1 Section 11.2.1
Supporting Document v1.1 Section 11.2.1 "The evaluator shall select one week at maximum"...  Does it matter how many TOE devices are being tested?  I assume one week means one person week (i.e., 40 hours) of effort, and maybe that should be clarified.  But the statement should also clarify how multiple test devices might affect the time frame.

woodbe:
This is to close #377.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -1749,7 +1749,7 @@ The evaluator shall take the following application notes into account to calcula
 
 ==== Application note for Elapsed time for Identification
 
-The evaluator shall select one week at maximum because the evalator shall finish the penetration testing within one week.
+The evaluator should select one week at maximum because the evalator should finish the penetration testing within one week for the testing of a single biometric sensor. This is not intended to force the evaluator to spend a week of time for testing each sensor type when multiple sensor types are tested in a single evaluation. The evaluator must be able to provide a sufficient justification that the time taken in testing for each sensor type is equivalent to a week for a single sensor test (when only a single sensor would be tested in isolation).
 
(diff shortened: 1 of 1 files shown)


Supporting Documents/BS_SD.adoc
```diff
@@ -1749,7 +1749,7 @@ The evaluator shall take the following application notes into account to calcula
 
 ==== Application note for Elapsed time for Identification
 
-The evaluator shall select one week at maximum because the evalator shall finish the penetration testing within one week.
+The evaluator should select one week at maximum because the evalator should finish the penetration testing within one week for the testing of a single biometric sensor. This is not intended to force the evaluator to spend a week of time for testing each sensor type when multiple sensor types are tested in a single evaluation. The evaluator must be able to provide a sufficient justification that the time taken in testing for each sensor type is equivalent to a week for a single sensor test (when only a single sensor would be tested in isolation).
```
gfiumara, line 1752: I understand the motivation, but want to make sure we don't make this _too_ lax. What do we think about this modification?

```suggestion
The evaluator should select one week at maximum because the evalator should finish the penetration testing within one week for the testing of a single biometric sensor. This is not intended to force the evaluator to spend a week of time for testing each sensor type when multiple sensor types are tested in a single evaluation. The evaluator shall provide a sufficient justification that the time taken in testing for each sensor type is equivalent to a week for a single sensor test (when only a single sensor would be tested in isolation).
```

n-kai (approved):

PR #387, opened 2022-06-10
No TSFIs output plaintext data

### LINKED ISSUE #371: Supporting Document v1.1 FPT_BDP_EXT.1 Section 2.4.1.7.1
Supporting Document v1.1 FPT_BDP_EXT.1 Section 2.4.1.7.1 "b. TSFI invocation test" - It is entirely unclear how an evaluator might "identify any TSFIs that output plaintext...". Also, the test concludes with essentially "if they cannot find any, no need to do this test" which seems backwards. This test should be recast and start with whatever the condition is (e.g., if the guidance has any TSFIs that..., then the evaluator shall test....).  However, it seems like this so-called test activity is really an ADV_FSP issue where the evaluator should be ensuring that none of the identified TSFIs indicate that they export plaintext data.

woodbe:
This is to close #371

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -570,7 +570,7 @@ In any case, the evaluator shall examine the TSS to confirm that;
 
 * If a biometric capture sensor returns plaintext biometric data, any entities outside the SEE can’t access the sensor and data captured by the sensor
 
-. All plaintext biometric data is retained in volatile memory within the SEE and any entities outside the SEE including the main computer operating system can’t access these data. Any TSFIs do not reveal plaintext biometric data to any entities outside the SEE
+. All plaintext biometric data is retained in volatile memory within the SEE and any entities outside the SEE including the main computer operating system can’t access these data. No TSFIs exist that reveal plaintext biometric data to any entities outside the SEE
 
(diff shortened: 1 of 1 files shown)


woodbe:
@n-kai I'm not sure that specific test works here readily. The TSFI could just write it to storage directly (for example from the SEE to a location outside the SEE), and the test you note here would miss that because it would not be in memory, but on disk. I think that the key needs to be stating that the TSFI needs to be documented and then how it is controlled (or blocked, or whatever) such that it cannot be accessed, and then when called that it does not generate whatever output it is supposed to do. Another example could be that the TSFI is an actual API that would just respond to the call directly (that could of course leave it in memory, but the real issue would be a calling app that knows about it would just receive the information without any need to scrape memory for it). I fear that getting too prescriptive in this test could miss possible methods for output that are more important than just a scrape of memory.

n-kai:
@woodbe 
I completely agree with _I think that the key needs to be stating that the TSFI needs to be documented and then how it is controlled (or blocked, or whatever) such that it cannot be accessed, and then when called that it does not generate whatever output it is supposed to do_. If the evaluator can get complete biometric sensor's API specification and how the dangerous API (e.g., dump the fingerprint image to the memory) are disabled, the evaluator should invoke the API and confirm that the API is disabled (and check no method to bypass the control of  API protection), rather than doing memory scan.

woodbe:
@woodbe to incorporate latest suggestions.

woodbe:
I am closing this PR in favor of #388 which is better tied to the #380 changes (and hence will be easier to merge since #388 is based on that PR).

PR #391, opened 2022-07-12
adding IAPAR definition

woodbe:
Added IAPAR definition in section 7 to respond to concerns about clarity of PAD testing.

This is related to biometricITC/cPP-toolboxes#58


Supporting Documents/BS_SD.adoc
```diff
@@ -858,7 +858,7 @@ The evaluator shall follow the same EAs defined in <<PP_MDF>> and the BIOSD. As
 
 === Introduction
 
-PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if any of these requirements are included.
+PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if any of these requirements are included. PAD is measured by using the Imposter Attack Presentation Attack Rate (IAPAR). IAPAR is a "full system" evaluation, and not a component evaluation, so the measure applies to the system as a whole, and so the results of failures for both PAD and the match are combined (the actual reason for the failure is not known, only that the presentation failed). According to <<ISO30107-3, ISO/IEC 30107-3>>, the proportion of presentation attacks using the same PAI species that results in an accepted verification is measured by IAPAR (Note that the 2017 edition defines this as Imposter Attack Presentation Match Rate (IAMPR) but this will be changed to IAPAR in a later revision).
```
gfiumara, line 861: Attempting to make this a footnote.

```suggestion
PAD is an optional set of requirements, when FIA_MBE_EXT.3 or FIA_MBV_EXT.3 are included in an evaluation. This set of Evaluation Activities are only performed if any of these requirements are included. PAD is measured by using the Imposter Attack Presentation Attack Rate (IAPAR). IAPAR is a "full system" evaluation, and not a component evaluation, so the measure applies to the system as a whole, and so the results of failures for both PAD and the match are combined (the actual reason for the failure is not known, only that the presentation failed). According to <<ISO30107-3, ISO/IEC 30107-3>>, the proportion of presentation attacks using the same PAI species that results in an accepted verification is measured by IAPARfootnote:[The 2017 edition defines this as Imposter Attack Presentation Match Rate (IAMPR) but this will be changed to IAPAR in a later revision.].
```

gregott (approved):

PR #392, opened 2022-08-18
date and allowed list

woodbe:
To match up with the PP_MDF release the following changes were made:

- release dates for the docs all changed to match with the dates of the PP_MDF date (September 12, 2022)
- The specific PP-Modules that are allowed was removed and replaced by a link to a new page on the website. This will prevent needing to revise the PPM every time NIAP publishes as new Module update as the website page can be done easily.
- in the PP-Config (which will not be used as NIAP will create their own), I changed the reference to the PPM to MOD_BIO_V1.0 which matches the NIAP format for the names

### DIFF
--- a/PP-Configurations/PP_MD/PP_Config.adoc
+++ b/PP-Configurations/PP_MD/PP_Config.adoc
@@ -3,7 +3,7 @@
 :toc:
 :table-caption: Table
 :revnumber: 1.1
-:revdate: July 19, 2022
+:revdate: September 12, 2022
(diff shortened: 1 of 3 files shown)


n-kai (changes requested):
Please merge #393 into Date & allowed list update branch.

n-kai (approved):

ccparran (approved):

gregott (approved):

gregott (approved):

Issue #416, opened 2023-08-22
PADv2 integration


woodbe:
1.
FIA_MBV_EXT.3.1
The TSF shall prevent use of artificial presentation attack instruments utilizing [selection: PADv1, PADv2] for the biometric modality.

FIA_MBV_EXT.3.2
The TSF shall provide a biometric verification mechanism with the IAPAR not exceeding [assignment: value equal to or less than 15% (15:100)] to prevent use of artificial presentation attack instruments from being successfully verified.

2.
Create a new SFR (based on the above FIA_MBV_EXT.3.1) that has a selection with [PADv1, PADv2, none] that is mandatory, and forces the selection.

The FIA_MBV_EXT.3 SFR becomes selection-based instead of optional

3.
FIA_PAD_EXT.1.1
The TSF shall prevent use of artificial presentation attack instruments utilizing [selection: PADv1, PADv2, no PAD] for the biometric modality for verification and [selection: enrolment, no other times].

4.
first select whether PAD is being done for verification, enrolment or none
second, if not none, then bring in PAD requirement as selection
PAD level can then be part of PAD requirement

FIA_PAD_EXT.1.1
The TSF shall prevent use of artificial presentation attack instruments for the biometric modality for [selection: verification, enrolment, no PAD].

woodbe:
To make PADv1 mandatory:

FIA_MBV_EXT.1.1
The TSF shall provide a biometric verification mechanism using [selection: 
- eye;
- face with FIA_MBV_EXT.3;
- fingerprint with FIA_MBV_EXT.3;
- vein
].

Using something like the first option from first comment.

woodbe:
First thought, keep PAD optional for now, so use this:

FIA_MBV_EXT.3.1
The TSF shall prevent use of artificial presentation attack instruments utilizing [selection: **PAD-L1, PAD-L2**] for the biometric modality.

FIA_MBV_EXT.3.2
The TSF shall provide a biometric verification mechanism with the IAPAR not exceeding [**selection: PAD-L1 [assignment: [with a value equal to or less than 15% (15:100)]], PAD-L2 [assignment: [with a value equal to or less than 10% (10:100)]]**] to prevent use of artificial presentation attack instruments from being successfully verified.

This then also increases the IAPAR requirements as the PAD level increases.

Then when making PAD-L1 mandatory, use something like from https://github.com/biometricITC/cPP-biometrics/issues/416#issuecomment-1688351788

[...]

PR #419, opened 2024-05-14
FIA_PAD_EXT

woodbe:
This is a proposed add for FIA_PAD_EXT. This doesn't complete all the linkage to the SPD and such that is needed, but I think it gets the SFR updates that are needed correct based on #416.


woodbe:
Needs to have tracing to the objectives verified as well as dependency tables.

Review all PAD descriptions to make sure this is mentioned properly (if needed).

n-kai (approved):
PAD L1 or L2 is not defined in the PPM. The CC requires that any technical terms in SFR should be clearly defined. Text for overview about PAD levels should be created (as proposed in #63 in cPP-toolboxes) and add reference to this text to the PPM. Term PAD L(evel) should also be added to 1.5 Glossary. But such modification should be done after creating an overview about PAD levels.

woodbe:
On looking at this integration again, and now that we are actually making plans for PAD-L3, I think maybe we should adjust how this was written to not have the level specifically tied to the modality. If we separate the PAD level from the modality then it is possible to have any PAD for any modality, but the lab/scheme, who would have to be members (or at least have access to the toolboxes( would know what is or isn't allowed and can enforce the restrictions that way without us needing to update the requirements again if/when we add more levels to more modalities.

At this point I don't see anything higher than PAD-L3 as likely, but if we had a selection for the PAD level (which could include PAD-L3 if we wanted, but knowing that this isn't acceptable at the moment), it should work.

woodbe:
I have made the update and included the PAD-L definition in the glossary

PR #421, opened 2024-10-23
PAD L2+ udpate

woodbe:
This change is to incorporate PAD-L3 into the SFRs and removing the specific enumeration of PAD for each modality.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -455,37 +455,11 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_PAD_EXT.1 Presentation attack support [[FIA_PAD_EXT.1]]
 
-*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _biometric modality_] with [*selection*: _enrolment detection_] detection at biometric enrolment and [*selection*: _verification detection equal to or greater than enrolment detection_] detection at biometric verification.
+*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _eye, face, fingerprint, vein_] with [*selection*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBE_EXT.3 during biometric enrolment and [*selection, equal or greater than enrolment*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBV_EXT.3 during biometric verification.
 
-*Application Note {counter:remark_count}*:: The <<PAD_Levels>> table provides acceptable options for each modality by row for each selection within the SFR. Selecting a specific modality then specifies the options that are available for enrolment and verification detection.
+*Application Note {counter:remark_count}*:: Not all combinations of modality and PAD levels may be supported as PAD-L2 and PAD-L3 toolboxes are not publicly released.
 
 *Application Note {counter:remark_count}*:: The selection for the verification detection shall be at least equal to the selection for the enrolment detection. Choosing PAD (either L1 or L2 as appropriate) at enrolment means that at least the same level of PAD shall be selected for verification. Higher levels for verification can always be selected (such as No PAD at enrolment and any PAD level for verification or PAD-L1 for enrolment and PAD-L2 at verification).
-
-.Supported PAD levels
-[[PAD_levels]]
-[cols=".^1,.^2,.^2",options=header]
-|===
-|Biometric modality
-|Enrolment detection
-|Verification detection
-
-|eye 
-|{empty}[*selection*: [.underline]#PAD-L1 as specified in FIA_MBE_EXT.3, no PAD#]
-|{empty}[*selection*: [.underline]#PAD-L1 as specified in FIA_MBV_EXT.3, no PAD#]
-
-|face
-|{empty}[*selection*: [.underline]#PAD-L2 as specified in FIA_MBE_EXT.3, PAD-L1 as specified in FIA_MBE_EXT.3, no PAD#]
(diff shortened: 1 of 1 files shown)


Protection Profile/BiocPP.adoc
```diff
@@ -455,38 +455,12 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_PAD_EXT.1 Presentation attack support [[FIA_PAD_EXT.1]]
 
-*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _biometric modality_] with [*selection*: _enrolment detection_] detection at biometric enrolment and [*selection*: _verification detection equal to or greater than enrolment detection_] detection at biometric verification.
+*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _eye, face, fingerprint, vein_] with [*selection*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBE_EXT.3 during biometric enrolment and [*selection, equal or greater than enrolment*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBV_EXT.3 during biometric verification.
 
-*Application Note {counter:remark_count}*:: The <<PAD_Levels>> table provides acceptable options for each modality by row for each selection within the SFR. Selecting a specific modality then specifies the options that are available for enrolment and verification detection.
+*Application Note {counter:remark_count}*:: Not all combinations of modality and PAD levels may be supported as PAD-L2 and PAD-L3 toolboxes are not publicly released.
```
gfiumara, line 460: I understand why this note is here and the new selection mechanism makes it more concise (details in [this comment](https://github.com/biometricITC/cPP-biometrics/pull/419#issuecomment-2433382642)). Two questions though:

 * We're not hiding the _existence_ of higher PAD testing levels, correct? The way the note is written might imply that we are.
 * Does CC allow for instructions/normative text in a note? To me this is essentially saying you need to go look up the existence of another document before you make this selection, in order to ensure that the selection value is valid. Having the table, while cumbersome, prevented an invalid choice.

woodbe: So honestly I wasn't exactly sure how much we wanted to state exactly which PAD-L2 toolboxes we have. I had sort of left it more "you need to ask". 

I will adjust though to incorporate this (and update the web page as well).

woodbe:
I have made the updates mentioned in the glossary as well as adjusting the website page to show the PAD-L2 fingerprint and the date it was published (but the link takes the user to the request access form).

Protection Profile/BiocPP.adoc
```diff
@@ -455,37 +457,11 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
 
 ===== FIA_PAD_EXT.1 Presentation attack support [[FIA_PAD_EXT.1]]
 
-*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _biometric modality_] with [*selection*: _enrolment detection_] detection at biometric enrolment and [*selection*: _verification detection equal to or greater than enrolment detection_] detection at biometric verification.
+*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _eye, face, fingerprint, vein_] with [*selection*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBE_EXT.3 during biometric enrolment and [*selection, equal or greater than enrolment*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBV_EXT.3 during biometric verification.
```
gfiumara, line 460:
```suggestion
*FIA_PAD_EXT.1.1*:: The TSF shall provide the [*selection*: _eye, face, fingerprint, vein_] with [*selection*: _PAD-L3_ (if available), _PAD-L2_ (if available), _PAD-L1_, _no PAD_] detection as specified in FIA_MBE_EXT.3 during biometric enrolment and [*selection, equal or greater than enrolment*: _PAD-L3, PAD-L2, PAD-L1, no PAD_] detection as specified in FIA_MBV_EXT.3 during biometric verification.
```

If you're allowed to use conditionals in selection (and it appears you can because we do it in *FIA_MBV_EXT.3.1*), then you could do something like this. Then the note is still explanatory and not normative. I think this helps resolve the issue I'm foreseeing.

woodbe: I can see where you are going, but it if we need to do that, then we should probably put it as the selection, like

[**selection, from available PAD levels for the modality**: _PAD-L3, PAD-L2, PAD-L1, no PAD_]

I don't really like the "if available" as part of the selection itself.

gfiumara: I would be fine with that. Let's perhaps get some other opinions on if this is necessary.

woodbe: First selection should be:

[**selection, from available PAD levels**: _PAD-L3, PAD-L2, PAD-L1, no PAD_]

Second selection should be:

[**selection, from available PAD levels, equal or greater than enrolment**: _PAD-L3, PAD-L2, PAD-L1, no PAD_]

Protection Profile/BiocPP.adoc
```diff
@@ -455,37 +457,11 @@ Target error rates defined in SFR shall be evaluated based on <<BIOSD>>. Normall
...
 
-*Application Note {counter:remark_count}*:: The <<PAD_Levels>> table provides acceptable options for each modality by row for each selection within the SFR. Selecting a specific modality then specifies the options that are available for enrolment and verification detection.
+*Application Note {counter:remark_count}*:: Not all combinations of modality and PAD levels may be supported.
```
woodbe, line 462: This should be updated to say that the website lists the available combinations of PAD and modality that are supported.

woodbe:
The implication of the change to this requirement as a selection is that the FIA_MBx_EXT.3 SFRs need to be moved to the selection-based section from optional.

Issue #398, opened 2022-10-11
[PPC CHANGE] version reference is incorrect


woodbe:
**What is the change request for the PP-Configuration? Please describe.**
[MOD_BIO_V1.0] is listed twice in the document. This should be [MOD_BIO_V1.1]

**Describe the solution you'd like**
replace 1.0 with 1.1

**Additional context**
This is not a high priority, but a typo that should be fixed at some point.

woodbe:
In the update change it to V2.0

PR #428, opened 2024-12-05
CC:2022 Update

woodbe:
This update targets the CC:2022 changes that are needed. NIAP pointed out that technically we can't publish a 3.1 PP (or module) anymore. So we can publish a CC:2022 update that still targets a 3.1 PP, but that PP would technically need to be evaluated to CC:2022 (which is, in this case, probably not an actual issue).

The thinking I have at the moment here is that we take any comments on the FIA_PAD_EXT.1 changes, wrap them with this and publish v2.0. When NIAP publishes the new PP_MDF, we can publish a 2.1 to match it (I'm assuming there may be some small changes beyond just changing the PP version, but I don't know).

Looking through all the changes from the DSC edits, we don't really have much to do here. The primary issues would be in changes to the requirements themselves, and here we have all EXT requirements, so there isn't anything there we need to take into account, and the requirements we have don't align with anything that is newly added to the catalog, so there isn't anything to cover.

As for the SD, again, since all the SFRs are EXT, the SD is what we write, and since we were following the normal methods of doing it, we don't really have anything to change.

I haven't set the version numbers in this to anything consistent yet, there are mainly some updated references to a 1.2, but it isn't consistently applied and would need to be checked across the board. Other than that though, this is probably complete given how the PP-Module is written.

### DIFF
--- a/PP-Configurations/PP_MD/PP_Config.adoc
+++ b/PP-Configurations/PP_MD/PP_Config.adoc
@@ -2,8 +2,8 @@
 :showtitle:
 :toc:
 :table-caption: Table
-:revnumber: 1.1
-:revdate: September 12, 2022
+:revnumber: 1.2
+:revdate: TBD
 :xrefstyle: full
 :doctype: book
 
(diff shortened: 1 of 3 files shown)


gregott (approved):

n-kai (approved):

woodbe:
This will close #398

PR #438, opened 2025-05-13
AGD guidance edit

### LINKED ISSUE #434: Correct use of AGD Evaluation Activities
**What is the change request for the BIOSD? Please describe.**
Section 2.3.5.4 in the SD (and possibly elsewhere) states "AGD guidance may include...." without defining an actual mandatory evaluation activity for the guidance document.

**Describe the solution you'd like**
Reword to include mandatory evaluation activity.

**Describe alternatives you've considered**
The guidance evaluation activity may be conditional or optional under certain circumstances and that is why it is structured that way; if so, this should be clearly explained through use of "conditional". 

**Additional context**
This may be present elsewhere in the document.

woodbe:
This is to close #434.

I am actually deleting the section that caused the problem as I think it's more of a copy-paste error now that I review the content. I think the rest do still need to be modified in their placement as they are sort of hanging on at the end. This change moves all of them into the requirements as part of the bullet.

This specific bullet though is about selecting PAD, and doesn't actually have any AGD associated with it because the user isn't going to turn it on or off, so it shouldn't be there.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -308,7 +308,7 @@ Following input is required from the developer.
 [loweralpha]
 . TSS shall explain how the TOE meets FIA_MBE_EXT.1 at high level description
 . AGD guidance shall provide clear instructions for a user to enrol to the biometric system
-
++
 AGD guidance may include online assistance, errors, prompts or warning provided by the TOE during the enrolment attempt.
 
 ===== Assessment Strategy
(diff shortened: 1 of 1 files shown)


gregott:
I think it makes more sense to have the two guidance statements together as you are doing now.  I have one suggestion: add an "s" to "warning", for e.g. "...online assistance, prompts or warnings provided by the TOE...".

woodbe:
@gregott I have made that change here.

gregott (approved):

n-kai (approved):

PR #440, opened 2025-06-30
AP Table Update

woodbe:
This updates the AP table and the text to the newer table. The one problem I found is that the table does not include the Window of Opportunity (Access to the TOE) which we need. The Degree of Scrutiny isn't sufficient for this use case.

I am proposing that as an addition to the table. It is possible they could be merged, but I don't think so off hand. They are sort of exclusive to each other.

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -1539,7 +1539,7 @@ Exploitation corresponds to achieving the attack on an instance of the TOE in it
 
 ==== Factors to be considered
 
-As in the [CEM], the factors to be considered consist of *_Elapsed time_*, *_Expertise_*, *_Knowledge of the TOE_*, *_Equipment_*, *_Window of opportunity_* and *_Degree of Scrutiny_*. But *_Window of opportunity_* is divided into two subfactors *_Window of opportunity (Access to the TOE)_* and *_Window of opportunity (Access to biometric characteristics)_*. 
+As in the <<CEM>>, the factors to be considered consist of *_Elapsed time_*, *_Expertise_*, *_Knowledge of the TOE_*, *_Equipment_*, *_Window of opportunity_* and *_Degree of Scrutiny_*. But *_Window of opportunity_* is divided into two subfactors *_Window of opportunity (Access to the TOE)_* and *_Window of opportunity (Access to biometric characteristics)_*. 
 
 *_Elapsed time_* is the total amount of time taken by the attacker.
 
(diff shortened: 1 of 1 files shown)


gregott:
I am not sure what you mean by "flexible support adaptation" in Knowledge of TOE > Equipment > Bespoke equipment (line 1571)

In the same paragraph, what is meant by "limited time access"?

gregott:
In Window of opportunity (Access to the TOE) > Non-cooperative, Do we want to add a sentence that the subject is not aware his biometric sample is being collected?   Also, what is the difference between "without full cooperation" and no cooperation?

woodbe:
@woodbe will rewrite limited time access in like 1571

woodbe:
@woodbe line 1587 replace "full" with "conscious" for cooperation.

woodbe:
@woodbe edit line 1601 - this should be no need to access the TOE, not the biometric characteristic

woodbe:
@woodbe  line 1608 fix authentication

woodbe:
@woodbe keep both Access to TOE and Degree of scrutiny. Make sure the explanatory text does not overlap given they are separate. Access to TOE descriptions will need to be adjusted.

woodbe:
@woodbe add proposed rows from ISO comments

woodbe:
@woodbe & @n-kai will look at "flexible support adaptation" text to see if we can come up with something better and clear.

n-kai:
> @woodbe & @n-kai will look at "flexible support adaptation" text to see if we can come up with something better and clear.

Attak potential chapter was developped based on the CC smart card supporting document (Application of Attack Potential to Smartcards and Simlar devices *1). I rephraised the para that includes "flexible support adaptation" and "limited time access" refering this document. 

_Bespoke equipment refers to very expensive equipment, the equipment is so specialized that its distribution is controlled, or the use of equipment requires either a very long experience or the human operator is hired together with the equipment; for example, the high-grade printing systems with the human operator hired together. In addition, if more than one specialized equipment is required to perform different parts of the attack, this value should be used. Before using this level, it shall be carefully checked the possibilities of renting equipment. If such service exists, the level shall be moved down to specialized level._

*1) https://www.commoncriteriaportal.org/files/ccfiles/JIL-Application-of-Attack-Potential-to-Smartcards-v3.2.1.pdf

gregott:
The rewording from @n-kai is great and makes the statements in question much more understandable.  Below is a version that is slightly "wordsmithed" for readability.

_Bespoke equipment refers to very expensive equipment. The equipment is so specialized that its distribution is controlled, or the use of equipment requires either significant experience to master its operation or a human operator is hired together with the equipment; for example, high-grade printing systems with a human operator hired together. In addition, if more than one specialized equipment is required to perform different parts of the attack, this value should be used. Before using this level, the possibility of renting equipment should be determined. If such a service exists, the level shall be moved down to the specialized level._

woodbe:
OK, I added the text from @gregott into the latest commit I just pushed.

woodbe:
I just updated the rating table to match what is in the proposed ISO document. There are some differences as we initially had AVA_VAN.1/2 as meeting the same thing whereas the updated table has those split out but has the AVA_VAN.4/5 as both being met at the =>40 level. I updated the table to reflect this change.

n-kai (changes requested):
See my pull request #441

n-kai (approved):
I will also update the table in ISO/IEC 19989-1 as described in this PR but the table in ISO may be modified/updated further based on comments from JWG7 experts. 
I will also add some examples of attack potential caluculation to ISO/IEC 19989-1 refering the attacks in the toolbox.

PR #442, opened 2025-08-27
TS 103 732-1 Integration

### LINKED ISSUE #427: ETSI TS 103 732-1 (CMD PP) as another Base-PP
The TS 103 732-1 PP will be updated to CC:2022 in 2025 and at that point I am expecting to try and point that to the PP-Module here. This will require more work.

1. Change overview to point to both Base-PPs (section 2.2)
2. create new section that is like 6.2 for the new Base-PP
3. New 8.4.1 section for consistency with new Base-PP
4. Consider how the low level requirements for the FPT_xxx SFRs are met in the new Base-PP

The real work I think will be in the SD though. The Base-PP is EAL2, which means we will have to justify how the requirements are met. Mainly this is section 6, but section 6.6 for AVA_VAN will be more work as EAL2 is AVA_VAN.2. This means that section 7.3 will also need updates.

woodbe:
This is integration for the TS 103 732-1 upcoming v3.1.1 update that will be CC:2022 compliant. This is based on the current draft I am working on in ETSI, but the requirements that are "changed" are unlikely to be modified as they are either CC:2022 new or existing from v2.1.2, so those changes should be safe as far as needing to be changed later.

For reference, the current published document is here: 

[https://www.etsi.org/deliver/etsi_ts/103700_103799/10373201/02.01.02_60/ts_10373201v020102p.pdf](https://www.etsi.org/deliver/etsi_ts/103700_103799/10373201/02.01.02_60/ts_10373201v020102p.pdf)

This still needs to have some AVA_VAN.2 info, but that is already being worked on separately, so I'm not going to include that here and expect that what we write in the other pull request will be sufficient.

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -468,14 +468,14 @@ that meets the following: [assignment: _list of standards_].
 ===== FIA_AFL.1 Authentication failure handling
 This SFR adds the specific biometrics allowed by the PP-Module and only makes changes to FIA_AFL.1.1.
 
-*FIA_AFL.1.1*:: The TSF shall detect when [an integer between 3 and 10] unsuccessful authentication attempts occur related to [selection: _password, PIN, pattern, *eye, face, fingerprint, vein,* EAF_].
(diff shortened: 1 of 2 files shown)


woodbe:
swap lines 738 & 849, they are backwards for which PP they are tied to

woodbe:
Edit lines 931 and 935, they reference MDF and not CMD requirements for testing

PR #443, opened 2025-09-11
AVA_VAN PP-Module updates

woodbe:
This is a pull request to add 2 new Annexes and a few other minor changes in the PP-Module to create the AVA_VAN.2/3 modules to support PAD L2/3.

BIOSD updates will probably be lighter, though there should be things about some of the AVA_VAN targets, though given that most of this is handled as part of the toolbox, it should be minor.


n-kai (commented):
I am not sure if I can understand the multi-assurance evaluation correctly and if L2PP-Module or L3PP-Module try to introduce this new paradigm. But if the answer is yes, you may replace A.6 with following text. But I think that we should also create the PP config for the multi-assurance evaluation. I think that LXPP-Module is the right way to go but not sure if the justification of removal of dependencies are acceptable.

**A.6. Security Assurance Requirements**
This L2PP-Module is created following the multi-assurance evaluation paradigm that consists in applying different assurance requirements to different parts of the TSF (sub-TSFs), while enforcing a global set of SARs for the entire TOE. In this case, the entire TOE corresponds to computer containing the biometric system and the sub-TSF corresponds to the biometric system.

The L2PP-Module augments the SARs from the PP-Module Base with the addition of AVA_VAN.2 to counter against T.Intentional_Attack that assumes higher attack potential for presentation attacks. The TOE evaluation should follow multi-assurance evaluation paradigm when the SARs for the TOE (i.e., computer) includes lower component, AVA_VAN.1. 

AVA_VAN.2 has additional dependencies on ADV_ARC.1, ADV_FSP.2 and ADV_TDS.1 in addition to those components depended by AVA_VAN.1, however, those dependencies are removed in this L2PP-Module because attack scenarios or recipes that require Enhanced-Basic attack potential are defined in Toolbox and the evaluator doesn’t need such design information to verify that the TOE can counter against T.Intentional_Attack.

woodbe:
So the issue I have with the PP-Configuration is that we don't actually own the Base-PP. What I think we should do is publish that separately instead of on the doc, showing how we would expect it to be used, and not as part of the document itself. 

I have made edits for this text in the L2 and L3, editing it accordingly. I left the last paragraph about the ST maybe already including the AVA_VAN, but that this is specific to PAD.

woodbe:
From 10/28 call

Add additional text justification on not needing design documentation due to the purpose of the higher AVA_VAN being focused on actual testing of the product and how it works against actual presentation attacks and not the design itself. As the subcomponent is expected to be part of a larger system that protects the biometric system itself from attack (at this point that would be the assumption anyway), actual presentation attacks are the method for bypassing it, and so should be the focus.

Design documentation is not required for this type of test as it is looking at the performance of how the system works as a whole, and so a zero-knowledge expectation is acceptable for the AVA_VAN level because the species being used for the test at the higher AVA_VAN is the main point, actually testing whether better species can bypass the system, regardless of the design.

PR #444, opened 2026-01-28
Add PP-Configs

woodbe:
PP-Configs that we will use as examples, not as official publications since we are just the PP-Module


n-kai (changes requested):
I created a PR to fix a few editorial mistakes (#464 ).

n-kai (approved):

PR #449, opened 2026-02-20
Update equipment section in AP table

n-kai:
I compared values in our AP table and ones in smartcard AP table (*1)

Both tables have the same factors, Elapsed time, Expertise, Knowledge of TOE and Equipment.
Values in Expertise and Knowledge of TOE for both table are the same but values in Equipment in smartcard AP table is "None (0 0)" and "Standard (1 2)"  but our table doesn't have "None" and start from "Standard (0 0)".   Our AP talbe should follow the smartcard table as much as possible if there is no clear reason.

*1) https://certification.enisa.europa.eu/document/download/836d47eb-13cf-4415-beb6-df106582f1c1_en?filename=EUCC_state_of_the_art_attack%20potential%20to%20smart%20cards%20%20v2%20%28final.pdf

### DIFF
--- a/Supporting Documents/BS_SD.adoc
+++ b/Supporting Documents/BS_SD.adoc
@@ -1719,21 +1719,25 @@ The levels are as follows:
 
 3+|*Equipment*
 
-|Standard 
+|None 
 |0 
 |0
 
+|Standard 
+|1 
+|2
+
 |Specialised 
-|2 
+|3 
 |4
 
 |Bespoke 
-|4 
+|5 
 |6
 
 |Multiple Bespoke
-|6
-|10
+|7
+|8
 
 |Not practical
 |*


woodbe:
So my issue here is that "none" makes no sense. How can I attack anything without equipment? I mean in the biometric world this would be the drive-by attack, so to speak, which is more about FAR/FRR than it is IAPAR, in my opinion. I mean if I can walk up to your phone and unlock the biometric by just pressing my finger to it (or showing my face), with no preparation or anything, then that doesn't seem to even need PAD, it seems more like either nothing is being done or that the FAR/FRR is inaccurate (for say a 1:50K unless I happen to somehow look like you or have close prints).

This makes me think about the issue raised by Somya last September that our exploitation values were too high (and no 12 hour point should be awarded, that the points for the exploitation phase of the testing should all go down by one, so a day is 1/2 instead of the current 1/3). I argued against the 12 hour drop, even if I'm not as certain about the rest (I could still see them all dropping a point even if we leave the 12 hour point). But here it seems like we are giving points to someone having a computer. I am not really sure that I even buy that for the smart card, really, especially given what is "none" even supposed to mean for equipment for the attack? How would I have any idea of an attack with nothing except the smart card in my hands? What am I supposed to be doing, taking it to the store and attempting the hack live? I don't get how that even works.

Looking at their tools table, I would a PC or workstation as "none" and probably do the same for software (given you could get a lot of those tools for free from open source). I just have a hard time giving a lot of points for attacks because someone owns a computer. Maybe we could use this but create a cost level for the "None" of maybe $1000 which would likely be largely consumed by the computer and maybe some software, leaving very little left for equipment, but still being what most would use. Then adding say a 3D printer would push you to "standard"?

I think in the end my fear is that because we are looking at "simple" attacks that have low attack potentials that suddenly don't make sense or are borderline because we have to give someone points for having a computer. Like all the vein attacks suddenly went to enhanced basic because they were all around 8-9 points before, and so now because a computer gets points, they become harder?

woodbe:
Oh, now I see it. They also have very different rating values on page 25. So we aren't comparing common things. So if we add 3 for standard equipment to our table, it blows us up, but if we add it like this, it doesn't. That said, it really makes the values tight. We would have to think about this for biometrics. They are, to an extent I think, trying to account for a higher floor of what you need and then a smaller grouping for actually exploiting things. I'm not sure if this works for us.

0-15 No rating
16-20 Basic
21-24 Enhanced-Basic
25-30 Moderate
31 and above High

n-kai:
Based on the table 14 (Rating of vulnerabilites and TOE resistance) on page 25, the maximum attack potential biometrics recognition can resist would be Enhanced-Basic or Moderate, as the BSI claims. I assume that Kevin created his attack potential table based on table 13 (Final table for the rating factors) but revised the table 13 so extensively in his document to scale the data injection attack potential up to 'High.' We previously agreed to follow his scenario— am I correct?

However, it seems Kevin has changed his mind; he’s now claiming that the latest data injection attacks should be classified as 'High' potential. I don't have any additional information on what led to this change.

woodbe:
Just an FYI to everyone. @KevinClabs confirmed that for the time being, they are continuing with the current AP tables, and not looking at the newer updates from JIL. This may be used in the next update after this one, but either way, the process for making those changes will be long, and changing them now would introduce the lag immediately, so instead, work will progress on the table from where it is now, so we will leave the table as it is.

n-kai:
I agree to use the current AP table in working branch and close this PR and will update my PRs to L1 vein, L1/L2 face toolbox to align with the current table.

PR #451, opened 2026-03-11
direct rationale changes

### LINKED ISSUE #450: Direct Rationale update
So the latest draft of the PP_MDF has moved to a Direct Rationale version which is explicitly allowed under CC:2022. The issue we have is that my understanding, based on my read of CC:2022 Part 1, is that we cannot use a traditional PP-Module with a Direct Rationale PP-Module Base.

This is largely handled by just removing the Security Objectives for the TOE and the related tables to only list assumptions and OEs and then map these directly to the requirements.

### LINKED ISSUE #458: Comment #4 Conformance to PP_MDF (Direct Rationale)
**COMMENT**

"The draft of MDF 4.0, while not currently explicitly stated, appears to be a direct rationale PP (no security objectives of the TOE were included).  Comments to this effect were made against the MDF 4.0 during the comment period.  

Section 5.1 of the MOD contains TOE SOs

CC2022 Part 1, Section 11.3.2.1 indicates:
A PP-Configuration:
— may be used in context with the Direct Rationale approach described in B.5 and C.2.3. In this case, all of the components of the PP-Configuration shall also use the Direct Rationale approach;  Note, C.2.3 states:  
PP-Modules can be written with the intention that they be used with components in their PP-Module Base that also use the Direct Rationale approach. In this case security objectives for the TOE are not included in the PP-Module and security objectives for the TOE's operational environment may be included."

**CHANGE**

The MOD will need to convert to a direct rationale MOD to be able to have MDF4.0 as its base, if the MDF 4.0 is, in fact, direct rationale.

woodbe:
This is to close #450 

So to support the direct rationale update to the PP_MDF, we have to be direct rationale as well. This is my pass at changing it.

The highlights of the changes are:
- Deleted OSP.Verification
- Created OSP.Error_Rates - this was to handle OSP.Verification as well as half of OSP.Enrol since that had two things in it, the idea being that this one requirement handles quality for enrolment and verification
- Change OSP.Enrol to OSP.Auth_Enrol which is specifically targeted to the requirement of the user authenticating before enrolment
- Created OE.Auth_Enrol to respond to OSP.Auth_Enrol
- Deleted all Security Objectives
- Adjusted the mapping tables to point from threats and OSPs to SFRs directly without the objectives in the middle
- Continued the relevant changes into the L2/L3 PP-Modules as well

This should work just fine. I only changed a few because given needing to already change everything, I tried to slightly tighten up the actual claims in terms of what they said (mainly the enrol one and then remove duplication in the error rates thing).

### DIFF
--- a/Protection Profile/BiocPP.adoc
+++ b/Protection Profile/BiocPP.adoc
@@ -304,14 +304,14 @@ The threat <<T.Scripted_Attack>> is defined as attacks which are scored as meeti
 
 === Organizational Security Policies
 
-[[OSP.Enrol]]OSP.Enrol::
-The TOE shall enrol a user for biometric verification, only after successful authentication of a user. The TOE shall ensure that templates are of sufficient quality in order to meet the relevant error rates for biometric verification.
+[[OSP.Auth_Enrol]]OSP.Auth_Enrol::
+The TOE shall only enrol a user for biometric verification after successful user authentication.
+
+[[OSP.Error_Rates]]OSP.Error_Rates::
+The TOE shall ensure that templates are of sufficient quality in order to meet the relevant error rates for biometric verification.
 
 [[OSP.Protection]]OSP.Protection::
 The TOE in cooperation with its environment shall protect itself, its configuration and biometric data.
-
-[[OSP.Verification_Error]]OSP.Verification_Error::
-The TOE shall meet relevant criteria for its security relevant error rates for biometric verification.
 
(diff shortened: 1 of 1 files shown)


n-kai (changes requested):
I created the #461 to modify your proposed text.

n-kai (changes requested):
See #463 to fix some errors. I don't have any other comments.

PR #463, opened 2026-03-30
Update conformance claims and security objectives

n-kai:
This is a patch to fix editorial errors (replace common criterria ... with [CCn])


woodbe (approved):
So this just changes everything to say CCn with a link (I like the link), but should we have the name written out AND have the link? Like "CC Part 1" that is linked? 

This is more about readability. I get that we have made the reference, and it's like PP_MDF, we don't write it out. I'm fine with it, I'm just asking.

n-kai:
Either option is acceptable, provided the following inconsistency is resolved.
- Errata and Interpretation, Version 1.**1**
- [CC-E&I] Errata and Interpretation for CC:2022 (Release 1) and CEM:2022 (Release 1), 001, Version 1.**2**, October 15, 2025.

woodbe:
Ahh, yeah, that would have been a typo.

If no one has an issue with this tomorrow (the change to CCn), we will merge it on the call. Otherwise I'll fix the other error.

